PR #2, opened 2020-08-27
Add setup payment page+cu beq0c0

jtaaa:
### What's Changed
I added the `SetupPayment` page and storybook. It uses the `updateStore` mutation to set the WiPay Account Number for a store.

### UI
Screen Shot 2020-08-27 at 10 10 07 AM

### Notes
After I add the `SetupDelivery` page I'm going to hook up all these pages in an Onboarding page that will include the Stepper and will smartly redirect to either the `CreateStore`, `SetupPayment` or `SetupDelivery` page or back to account page depending on what is already set up for the store.


jtaaa:
Task linked: [CU-beq0c0 Add SetupPayment page](https://app.clickup.com/t/beq0c0)

jtaaa:
The changes in the following files are automatically generated (not necessary for review):
- packages/seller-client/graphql.schema.json
- packages/seller-client/src/generated/graphql.tsx
- packages/server/schema.graphql

packages/seller-client/src/components/Spacer/Spacer.tsx
```diff
@@ -2,10 +2,12 @@ import { styled, Theme } from '@material-ui/core/styles';
 
 type SpacerProps = {
   width?: number;
+  height?: number;
 };
 const Spacer = styled('span')<Theme, SpacerProps>({
   display: 'inline-block',
-  width: ({ width = 8 }) => `${width}px`,
+  width: ({ width = 1 }) => `${width}px`,
+  height: ({ height = 1 }) => `${height}px`,
```
chriskrogh, line 10: how come `${x}px` is required here instead of `x`?

packages/seller-client/src/pages/OnboardingPage/SetupPayments/PaymentsForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+          type="number"
+          label="WiPay Account Number"
+          variant="outlined"
+          fullWidth
+          inputRef={register({
+            required: 'Required',
+            minLength: {
+              value: 10,
+              message: 'WiPay account numbers must contain 10 digits',
+            },
```
chriskrogh, line 60: smol nit: Is there a way to do easily max length as well? If it's always 10 then we probs wanna stop them at 10. 

I see that we handle the case where the backend req fails by showing an error message so we don't _have_ to check max length here, but if possible that would be cool.

packages/server/src/resolvers/Store/inputs.ts
```diff
@@ -39,7 +39,10 @@ export class UpdateStoreInput implements Partial<Store> {
   description?: string;
 
   @GraphQLField({ nullable: true })
-  email?: string;
+  phoneNumber?: string;
```
chriskrogh, line 42: What's the reason behind this change?

packages/seller-client/src/components/Spacer/Spacer.tsx
jtaaa: it's not! good catch!

packages/seller-client/src/pages/OnboardingPage/SetupPayments/PaymentsForm.tsx
jtaaa: Yep and I fully intended to but forgot! good catch!

packages/server/src/resolvers/Store/inputs.ts
jtaaa: I needed to add wipayAccountNumber and noticed phoneNumber was missing. Technically we could allow sellers to update their store email address since authentication is linked to their User. The user email can't change because it could be provided by an auth provider (Google or FB).

At the time I was thinking it would require some extra steps to change your email as a store but it might not be an issue at all. I'll add it back

PR #3, opened 2020-08-28
[SetupDelivery - P1] Add initial SetupDelivery component+cu bmh8au

jtaaa:
### What's Changed

Added the initial `SetupDelivery` component and the `SetupPickupCard` component.

Screen Recording 2020-08-28 at 6 17 22 AM

### Notes

This feature differs from the designs quite a bit! I think it's an upgrade 💯 

### Up Next

Next I'll be adding the Delivery `SlideTabs` below the pickup stuff. Possibly on a card too!

---

Linked to CU-bmh8au


jtaaa:

jtaaa:
Link to CU-bmh8au

packages/seller-client/src/pages/OnboardingPage/SetupDelivery/SetupDelivery.tsx
```diff
@@ -0,0 +1,53 @@
...
+import Link from 'components/Link';
+import SetupPickupCard from './SetupPickupCard';
+
+const SetupDelivery: React.FC = () => {
+  const { data } = usePickupQuery();
+  const isEnabled = !!data?.store?.pickup?.enabled;
+
+  const { addressOptions, addAddress, loading } = useStoreAddresses();
+  const [disablePickup] = useDisablePickupMutation();
+  const [enablePickup] = useEnablePickupMutation();
```
chriskrogh, line 24: It makes sense that these 2 are `async`. But would there be a performance benefit if we used local state to manage these from the UI side and then when the user "submits" then we reach out to the back end?
Just curious because if someone's internet was pretty slow it would take a while to see the UI change reflected? Or is this stuff all cached?
I'm actually not sure if this is a big deal or if this is the convention that you're been following so far. My only concern is users clicking on something and having to wait a while for a response. They could think that our app is slow.

packages/seller-client/src/pages/OnboardingPage/SetupDelivery/SetupDelivery.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { DocsLinks } from 'utils/links';
+import { useStoreAddresses } from 'components/SelectAddress';
+import LogoHeader from 'components/LogoHeader';
+import Spacer from 'components/Spacer';
+import Link from 'components/Link';
+import SetupPickupCard from './SetupPickupCard';
+
+const SetupDelivery: React.FC = () => {
+  const { data } = usePickupQuery();
+  const isEnabled = !!data?.store?.pickup?.enabled;
```
chriskrogh, line 20: Just curious, what is the relationship between client and server? Is it kind of like `mern` where the server serves up the client using a proxy and exposes the route for graphql? Or do the client and server run on different ports?

chriskrogh:
> Link to CU-bmh8au

What's this?

packages/seller-client/src/pages/OnboardingPage/SetupDelivery/SetupDelivery.tsx
jtaaa: Great point! yep this could definitely make the app feel slower. Apollo client has a pattern called [Optimistic UI](https://www.apollographql.com/docs/react/performance/optimistic-ui/) but honestly it seems like overkill here.

Instead I'll refactor to run enablePickup on submit in my next PR

This also raises the point that we don't really have any loading indication on submit for any of these forms 😓

packages/seller-client/src/pages/OnboardingPage/SetupDelivery/SetupDelivery.tsx
jtaaa: I have it setup that the seller-client runs on `localhost:3001` locally and the Express server is run on `localhost:3000` with our Apollo server (which is run from within the Express server) handling routes on `localhost:3000/graphql`.

The marketplace-client NextJS app is served by the Express server (almost like a `mern` app but with SSR) on the root server endpoint (localhost:3000). The seller-client is not served by the server so we can take advantage of caching since it's a static SPA.

I plan on deploying to seller-client to seller.sellyourstuff.online and the server to sellyourstuff.online

jtaaa:
> > Link to CU-bmh8au
> 
> What's this?

I tried to link to ClickUp but it doesn't support linking in comments I guess :(

packages/seller-client/src/pages/OnboardingPage/SetupDelivery/SetupDelivery.tsx
jtaaa: CORS is currently enabled for `['http://localhost:3001', 'http://localhost:9009']`

chriskrogh: Ahh that's pretty smart. Makes sense 2 me!

PR #4, opened 2020-08-28
[SetupDelivery - P2] Add SetupDelivery component p2+cu beq0rt

jtaaa:
### What's Changed

Added the `SYSDeliveryForm` and `CustomDeliveryForm` to the `SetupDelivery` component in a `SlideTabs` component.

Screen Recording 2020-08-28 at 8 31 35 AM

Also added alert to display any API errors when setting delivery type. The API shouldn't through an error here since we validate on the client side before submitting the GraphQL mutations but just in case we have an alert with the error message printed. Maybe it could fail due to network errors or something else.

### Next Up

Next is creating the `OnboardingPage` with the stepper!


jtaaa:

PR #5, opened 2020-08-28
[SetupDelivery - P3] Enable pickup only on submit+cu beq0rt

jtaaa:
### What's Changed

We now only actually enable pickup in the backend only once the user clicks the "USE ... DELIVERY" button.

Also I've removed the queries in favour of having a default for onboarding. At this point in onboarding the user shouldn't have any values set so no need to query.


jtaaa:

packages/shared/src/types.ts
```diff
@@ -0,0 +1 @@
+export type Awaitable<T> = T | PromiseLike<T>;
```
chriskrogh, line 1: Pretty cool. Will remember to use

PR #7, opened 2020-08-29
[Cu-bmhppu - p2] Change return type of createStore mutation

jtaaa:
### What's changed

In order to take advantage of the automatic cache updating capabilities of `apollo-client` we must return the updated `User` document along with the store for the `createStore` mutation.

#### Fixes
- Made the `wipayAccountNumber` field nullable to match the type.


jtaaa:
Task linked: [CU-bmhppu Add Onboarding Page](https://app.clickup.com/t/bmhppu)

packages/server/schema.graphql
```diff
@@ -248,7 +253,7 @@ type Store {
   pickup: Pickup
   slug: String!
   user: User!
-  wipayAccountNumber: String!
+  wipayAccountNumber: String
```
chriskrogh, line 256: What do these ! do?

jtaaa:
> What do these ! do? 

It makes the type before it non-nullable. So in this case `wipayAccountNumber` can be `null` in a `Store` object type but `user` cannot. Apollo server will send an error instead of sending null for non-nullable fields.

PR #11, opened 2020-08-29
[CU-bmhppu - P4] Update useStoreAddress for caching

jtaaa:
### What's Changed

We're taking care of the apollo cache properly with update functions in our custom hook wrapping the one generated by apollo-codegen. I also added the generated files in this PR.

### Notes

In the future we can look into using Optimistic UI and improve performance!


jtaaa:
Task linked: [CU-bmhppu Add Onboarding Page](https://app.clickup.com/t/bmhppu)

jtaaa:
```ts
  const [addStoreAddress] = useAddStoreAddressMutation({
    update(cache, { data }) {
      cache.modify({
        fields: {
          // ...
        },
      });
      // ...
    }
});

```

We're updating the cache with an update method property on the options that we pass into `useAddStoreAddressMutation`. This is es6 method definition syntax equivalent to

```ts
 const [addStoreAddress] = useAddStoreAddressMutation({
    update: (cache, { data }) => {
      cache.modify({
        fields: {
          // ...
        },
      });
      // ...
    }
});
```

Each field is a function (again using method definition syntax) with the name being a field in the Apollo cache. The function takes the existing items or item in the cache (refs only) and returns what the new items in the cache should be.

```ts
          addresses(existingAddressRefs: Address[] = []) {
            const newAddressRef = cache.writeFragment({
              data: data?.addStoreAddress,
              fragment: AddressOptionFragmentDoc,
              fragmentName: 'AddressOptionFragment',
            });
            return [...existingAddressRefs, newAddressRef];
          },
```

The things that are in the cache are refs to fragments or objects. To add a new fragment (piece of an object) or full object to the cache you write a fragment using:

1. the data to be used
2. the fragment being added
3. if you have multiple fragments you have to include the fragment name (I think this is a limitation of `apollo-client`)

In case you're interested!

PR #25, opened 2020-09-01
Created RelativeDate Component

chriskrogh:
## What's Changed
Created RelativeDate Component

If date is 
today -> `Today`
within the next 7 days -> `Thu Sep 03 2020`
anytime further away -> `dd/mm/yyyy`

If a prefix is present, then we prepend it and **bold** the date text. Otherwise the date text has normal weight.
I noticed on Figma that the Typography `variant` was `body2` so I used that.

## UI
Today, prefix
Screen Shot 2020-09-01 at 12 16 07 PM
Within a week, prefix
Screen Shot 2020-09-02 at 1 19 41 PM
Further away, prefix
Screen Shot 2020-09-02 at 1 20 30 PM
Icon, Today, prefix
Screen Shot 2020-09-02 at 1 20 45 PM


jtaaa:
Task linked: [CU-bcguq4 Add RelativeDate component](https://app.clickup.com/t/bcguq4)

packages/seller-client/src/components/RelativeDate/RelativeDate.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+import RelativeDate from './RelativeDate';
+
+export default {
+  title: 'Components/RelativeDate',
+  component: RelativeDate,
+};
+
+const today = new Date();
+const WithinAWeek = add(today, { days: 2 });
+const furtherAway = add(today, { days: 10 });
```
jtaaa, line 12: Love these 😍

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  if (distanceToNow.includes('minute') || distanceToNow.includes('hour')) {
+    return 'Today';
+  } else if (
+    distanceToNow.includes('day') &&
+    +distanceToNow.split(' ')[0] < 7
+  ) {
+    return date.toDateString();
+  }
+  return `${date.getDate()}/${date.getMonth()}/${date.getFullYear()}`;
+};
```
jtaaa, line 37: Checkout [`date-fns/formatRelative`](https://date-fns.org/v2.16.1/docs/formatRelative)

This works so I won't block based on this! and the fact that it's a helper function means we can always update it easily to change it's implementation

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <Box display="flex">
+      {showIcon && <TodayIcon className={classes.icon} />}
+      {prefix && <Typography variant="body2">{prefix}</Typography>}
+      <Typography
+        variant="body2"
+        className={classnames(classes.date, prefix && classes.bold)}
+      >
+        {formatDate(date)}
+      </Typography>
```
jtaaa, line 50: The designs call for this text to be `variant="subtitle2"`! Let me know if you're having trouble viewing the text variants in Figma

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const classes = useStyles();
+  return (
+    <Box display="flex">
+      {showIcon && <TodayIcon className={classes.icon} />}
+      {prefix && <Typography variant="body2">{prefix}</Typography>}
+      <Typography
+        variant="body2"
+        className={classnames(classes.date, prefix && classes.bold)}
+      >
+        {formatDate(date)}
```
jtaaa, line 49: Can we have the date in `<strong>...</strong>` tags to make it bold if and only if a prefix is passed?

jtaaa: Ahh okay I see that the `fontWeight: bold` style is applied the class. I really much rather use the `strong` tag but I guess it's just preference.

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+  return `${date.getDate()}/${date.getMonth()}/${date.getFullYear()}`;
+};
+
+const RelativeDate: React.FC<Props> = ({ showIcon, prefix, date }) => {
+  const classes = useStyles();
+  return (
+    <Box display="flex">
+      {showIcon && <TodayIcon className={classes.icon} />}
+      {prefix && <Typography variant="body2">{prefix}</Typography>}
```
jtaaa, line 44: It might be worthwhile to put the icon and prefix inside the main Typography component so everything shares the same styles. The icon will have a size matching the text based on the `fontSize` of the Typography element.
You can use `<strong>` for bolding or apply the bold class to a span inside the Typography component.

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
chriskrogh: Nope I can see em! I must have missed this one, will pay closer attn next time 🎯
Thanks 4 the catch

packages/seller-client/src/components/RelativeDate/RelativeDate.stories.tsx
chriskrogh: tq!

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
chriskrogh: Ah good to know! We can stick with the custom helper since it's already written, but will keep in mind for future implementation!

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
chriskrogh: updated in 340d2305fbc3f2d4363087e1583bf6f2bac594ce

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
chriskrogh: I don't mind hopping on the 💪 train. updated in b48f5927fc5aabe205eef8c47afb61aab4a77218

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
chriskrogh: Updated to use a single typography component in 0998769d0af1526eecf0805a40acd4f956f0bf52
But I didn't include the icon in there because the alignment was really bad and I didn't know how to fix it.

Screen Shot 2020-09-02 at 1 15 35 PM

Typically I'd make this `display: flex` and center the content using `justifyContent` but that seems more like a job for a `Box` and not a `Typography`. How would you resolve that issue? Or do you think it's okay the way it is now with the prefix and date in 1 Typography and the Icon separate?

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { formatDistanceToNow } from 'date-fns';
+import { Typography, Box } from '@material-ui/core';
+import TodayIcon from '@material-ui/icons/Today';
```
jtaaa, line 5: Btw you can import like
```
import { Today } from '@material-ui/icons';
```
safely. I've set up tree shaking with `babel-plugin-import` so bundle size will still be optimized 🚀

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
```diff
@@ -0,0 +1,52 @@
...
+const RelativeDate: React.FC<Props> = ({ showIcon, prefix, date }) => {
+  const classes = useStyles();
+  const formattedDate = formatDate(date);
+  return (
+    <Box display="flex" alignItems="center">
+      {showIcon && <TodayIcon className={classes.icon} />}
+      <Typography variant="subtitle2">
+        {prefix ? (
+          <>
+            {`${prefix} `}
```
jtaaa, line 41: I think this way is a little more direct

```suggestion
            {prefix}{' '}
```

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
jtaaa: Fine by me! I expected the icon to be well positioned 😢 Looks good how it is now

packages/seller-client/src/components/RelativeDate/RelativeDate.tsx
chriskrogh: Sick! Will keep in mind

PR #26, opened 2020-09-01
Add TagsInput component

jtaaa:
### What's Changed

I added the `TagsInput` component which will be used to add `Tag`s to products. It works like the `SelectAddress` component in that it takes an `addTag` function that will create new tags in the backend and add it to the array passed to the `tags` prop which contains `TagOptions`.

Screen Recording 2020-09-01 at 4 50 21 PM

I also added **unit tests** 😍  for this component. Ideally all our components should have unit tests but because we are on a really tight schedule for v1, we will have to survive for now with some unit tests being skipped.

### Notes

I created a single unit test with multiple checks inside. I wasn't expecting this test to get so big so maybe I should have split it up into multiple tests.

In general we want to avoid having too many tests for one component by merging multiple checks together in one test to keep testing time short since we need to render the dom less times. The tests should be from the perspective of how a user would normally use the component and so **should** include multiple checks. We gotta find a balance between manageable test sizes and number of tests.

This PR might look like a big fella but a lot of it is the `yarn.lock` and the tests! Hopefully it should be too terrible to review! Let me know if it is 👂


jtaaa:
Task linked: [CU-bmqrec Add TagsInput component](https://app.clickup.com/t/bmqrec)

packages/shared/src/string.ts
```diff
@@ -14,3 +14,10 @@ export const joinWithAnd = (strs: string[]): string => {
   const last = strs[strs.length - 1];
   return lead.length > 0 ? `${joinedLead} and ${last}` : last;
 };
+
+export const toTitleCase = (str: string): string =>
+  str
+    .split(' ')
+    .filter((word) => !!word[0])
+    .map((word) => word[0].toUpperCase() + word.slice(1).toLowerCase())
+    .join(' ');
```
chriskrogh, line 23: Good to have!

packages/seller-client/src/components/TagsInput/TagsInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+export type TagsInputProps = {
+  tags: TagOption[];
+  selected?: TagOption[];
+  defaultSelected?: TagOption[];
+  onChange(newTags: TagOption[]): Awaitable<void>;
+  addTag(data: NewTagInput): Awaitable<TagOption>;
+  error?: boolean;
+  helperText?: string;
+};
+const TagsInput: React.FC<TagsInputProps> = ({
```
chriskrogh, line 21: OCD nit: it would be nice to have a `\n` between these 2 boys, but this has no impact on anything so feel free to ignore this

packages/seller-client/src/components/TagsInput/TagsInput.test.tsx
```diff
@@ -0,0 +1,124 @@
+import React from 'react';
+import { render, screen, waitFor, fireEvent } from '@testing-library/react';
+import user from '@testing-library/user-event';
+
+import type { TagOption } from './types';
+
+import { Base } from './TagsInput.stories';
+import { MockTags } from './mocks';
+
+test('Selects tags from provided list and adds new tags', async () => {
```
chriskrogh, line 10: Yea this is a pretty big boy. It might be a bit tricky to know what went wrong if he fails, but as of right now it's good to have em. I'm not too picky about unit tests

packages/seller-client/src/components/TagsInput/TagsInput.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { useTagsInputMocks } from './mocks';
+
+export default {
+  title: 'Components/TagsInput',
+  component: TagsInput,
+};
+
+type BaseProps = Partial<
+  Record<keyof Pick<TagsInputProps, 'onChange' | 'addTag'>, jest.Mock>
+>;
```
jtaaa, line 16: Allows us to pass our mock functions for our tests which will use the implementations from `useTagsInputMocks`.

packages/seller-client/src/components/TagsInput/TagsInput.tsx
jtaaa: Hahaha ok I'll add newlines from now on! I usually don't but doesn't hurt too + will fix ur OCD

packages/seller-client/src/components/TagsInput/TagsInput.test.tsx
jtaaa: Yeah it would be tricky to know what went wrong. I'll ticket a low-priority refactor.
I need to get back into testing to find the sweet spot for test size.

jtaaa:
Test refactor ticketed https://app.clickup.com/t/buha8q

PR #27, opened 2020-09-02
Add OrderModel

jtaaa:
### What's Changed

We now have an `OrderModel` that defines our `Order` mongoose schema and our `Order` GraphQL Object Type.

I also added an initial `OrderResolver` to test the GraphQL schema since Object Types are only added to the schema if there is a resolver of that Object Type added to the schema (or maybe also if a resolver includes the Object Type).

@chriskrogh you should be able to define a fragment in a `<name>.graphql` file that declares all the props needed for the `OrderListItem` component. You don't need to "hook up" the `OrderListItem`. You can just import that generated fragment and use it as a type!

### Notes

The `sysFee` should be set based on the store's tier. Right now the `tier` property is stored on the `UserModel` but it should really be on the `StoreModel`. I'll ticket and make another PR to fix that. [CU-buk3dy](https://app.clickup.com/t/buk3dy)

**A note on testing**: I haven't tested writing a doc to the DB using the schema generated by Typegoose from the `OrderModel`. I'll make sure everything is working when I implement the required resolver mutations. 

**items**: The `items` prop should not contain refs to the existing `ProductOption`s. Instead it should contain copies of the `ProductOption`s (and their respective products) at the time of order creation. This way if a `ProductOption` is update or deleted later it does not affect our records. I'm still thinking about how to handle this boy but I'll keep the GraphQL types the same regardless of implementation. So the current GraphQL types are final (or close to final). [CU-buk45g](https://app.clickup.com/t/buk45g)

CU-7grryj[merged]


jtaaa:
Task linked: [CU-7grryj Add Orders Model](https://app.clickup.com/t/7grryj)

PR #28, opened 2020-09-02
Created StatusStepper Component

chriskrogh:
## What's Changed
Created StatusStepper Component.
This components takes 
1. an array of `Status: { label: string, Icon?: icon component }` because we use icons instead of numbers for certain steps
2. an `activeStepNumber`. We use this to highlight the active step

I also created a `Step` component which takes 
1. a single `Status`
2. `stepNumber: number`
3. `isActive?: boolean`

just to split the component.

Something that I also want your opinion on is the way that we render the steps and dividers. Because we have horizontal dividers between steps, it was actually kind of tricky to get things centred and spaced evenly by having each step render a divider. That's because the actual content would be left aligned and we want the divider to be evenly in between the content. 

Eventually I found it to be easier to map over an array that had `null` between `Status` just because we can render dividers evenly among the rest of the content when we see `null`. 
To me it feels good to have the `display: flex, justifyContent: space-between` handle all of the spacing and not have to pad / margin the dividers but lmk what you think

## UI
Screen Shot 2020-09-02 at 11 52 52 AM

Screen Shot 2020-09-02 at 11 53 20 AM


jtaaa:
Task linked: [CU-bcgt8b Add Order Status Stepper](https://app.clickup.com/t/bcgt8b)

packages/seller-client/src/utils/theme.ts
```diff
@@ -36,6 +36,7 @@ let theme = createMuiTheme({
     },
     text: {
       primary: grey[800],
+      secondary: 'rgb(0, 0, 0, 0.38)',
```
chriskrogh, line 39: Messaged on slack about this. I'm not sure if this is the best place to store this color. We definitely want it somewhere on theme because we use it to represent `inactive` text, but I'm not sure. Open 2 feedback

packages/seller-client/src/components/StatusStepper/StatusStepper.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const classes = useStyles();
+  const spacedStatuses = insertNullBetweenElements(statuses);
+  return (
+    <Box display="flex" justifyContent="space-between">
+      {spacedStatuses.map((element, index) =>
+        element ? (
+          <Step
+            status={element}
+            stepNumber={getStepNumber(index)}
+            isActive={getStepNumber(index) === activeStepNumber}
```
chriskrogh, line 45: I'm not a huge fan of `getStepNumber` and calling it twice. Ideally, we'd just make `index / 2 + 1` a const somewhere in the map and then return the element, but I don't think I can do that?

jtaaa, line 45: Regarding steps and dividers we can avoid the `insertNullBetweenElements` by doing something like this:

```ts
return (
  <Box display="flex" justifyContent="space-between">
    {statuses.map((element, index) =>
      <>
        {index !== 0 && <Divider />}
        <Step ... />
      </>
    )}
  </Box>
)
```

And as for the `getStepNumber` q, we can 100% give this anonymous function a function body

```ts
return (
  <Box display="flex" justifyContent="space-between">
    {statuses.map((element, index) => {
      const stepNumber = getStepNumber(index);
      return (
        <>
          {index !== 0 && <Divider />}
          <Step
            status={element}
            stepNumber={stepNumber}
            isActive={stepNumber === activeStepNumber}
          />
        </>
      );
    })}
  </Box>
)
```

packages/seller-client/src/components/StatusStepper/Step.tsx
```diff
@@ -0,0 +1,60 @@
...
+        variant="body1"
+        color={isActive ? 'textPrimary' : 'textSecondary'}
+      >
+        <strong>{label}</strong>
+      </Typography>
+    </Box>
+  );
+};
+
+export default withStyles(styles, { withTheme: true })(Status);
```
jtaaa, line 60: Why not use `useTheme` and `useStyles`? h**o**o**k**s

packages/seller-client/src/components/StatusStepper/types.ts
```diff
@@ -0,0 +1,7 @@
+import type { OverridableComponent } from '@material-ui/core/OverridableComponent';
+import type { SvgIconTypeMap } from '@material-ui/core/SvgIcon';
+
+export type Status = {
+  label: string;
+  Icon?: OverridableComponent<SvgIconTypeMap>;
```
jtaaa, line 6: What does this do?

packages/seller-client/src/components/StatusStepper/Step.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const color = isActive
+    ? theme.palette.secondary.main
+    : theme.palette.text.secondary;
+  return (
+    <Box display="flex" flexDirection="column" alignItems="center">
+      <Box
+        className={classes.stepNumber}
+        style={!Icon ? { backgroundColor: color } : {}}
+      >
+        {Icon ? <Icon style={{ color }} /> : stepNumber}
```
jtaaa, line 48: I think `stepNumber` should be in a `Typography` component

packages/seller-client/src/utils/theme.ts
jtaaa: Turns out the [default theme](https://material-ui.com/customization/default-theme/) already has `theme.palette.text.disabled` and `theme.palette.text.hint` that are exactly `rgb(0, 0, 0, 0.38)`. I would prefer to add a `theme.palette.text.inactive` so that we can customize inactive styles separately from disabled or hint.

I made a ticket for this https://app.clickup.com/t/burady

jtaaa (commented):
Styles look really good in the screenshots 💯 

I've got some suggestions

packages/seller-client/src/components/StatusStepper/StatusStepper.tsx
jtaaa: But if we get rid of `insertNullBetweenElements` we don't even need to use `getStepNumber` since the index (or index + 1) is the step number.

jtaaa: Answered here https://github.com/SYSTT/sys-monorepo/pull/28#discussion_r482382451

packages/seller-client/src/utils/theme.ts
chriskrogh: Removed from theme in c7f6369423e53e046590cde1ae9c239605afc3d7 and used `theme.palette.text.disabled` 
I'll also assign https://app.clickup.com/t/burady to me so that I can make the change super quick

packages/seller-client/src/components/StatusStepper/StatusStepper.tsx
chriskrogh: Thanks!

packages/seller-client/src/components/StatusStepper/Step.tsx
chriskrogh: Tossed em in a Typo component in a86c093eb359f4dc58b1ec403c4d87f05d0e7f18

packages/seller-client/src/components/StatusStepper/types.ts
chriskrogh: That's the type of any Icon imported from `@material-ui/icons`, I guess using this type does limit us to only use `@material-ui/icons` in the `StatusStepper`. Do you anticipate us using icons from elsewhere anytime soon? or do you think we can use `@material-ui/icons` for now and refactor later if we ever need to use icons from elsewhere?

packages/seller-client/src/components/StatusStepper/StatusStepper.tsx
chriskrogh: Yea what was blocking me was the fact that I had the `Step` and the `Divider` in a `Box`  and not in `<></>`. This makes it so much easier. Updated in 1c226acdb22b31d5df646cb9f921cc61753e6180

packages/seller-client/src/components/StatusStepper/types.ts
jtaaa: Nope I'm fine with using `@material-ui/icons` as much as possible for now. Didn't realize this is the way they are typed.

packages/seller-client/src/components/StatusStepper/Step.tsx
```diff
@@ -0,0 +1,58 @@
...
+  stepNumber: {
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+    width: STEP_SIZE,
+    height: STEP_SIZE,
+    borderRadius: STEP_SIZE / 2,
+    color: 'white',
+    marginBottom: spacing(1),
+  },
```
jtaaa, line 19: Just btw all this could be done in `<Box />` if you wanted to!

packages/seller-client/src/components/StatusStepper/Step.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const { label, Icon } = status;
+  const color = isActive
+    ? theme.palette.secondary.main
+    : theme.palette.text.disabled;
+  return (
+    <Box display="flex" flexDirection="column" alignItems="center">
+      <Box
+        className={classes.stepNumber}
+        style={!Icon ? { backgroundColor: color } : {}}
```
jtaaa, line 40: This can be done with `bgcolor={!Icon ? color : 'initial'}` or something similar

packages/seller-client/src/components/StatusStepper/Step.tsx
```diff
@@ -0,0 +1,58 @@
...
+      >
+        {Icon ? (
+          <Icon style={{ color }} />
```
jtaaa, line 43: Can probably be achieved through the `color` prop when we add inactive to the theme? We'll see!

jtaaa (approved):
LGTM

I commented suggestions that are 100% not necessary but maybe good to know 🚀

packages/seller-client/src/components/StatusStepper/Step.tsx
chriskrogh: Box is the boss yes! updated in 93d4d86249d32b33af3161b8ed37cc3ab96137fb

packages/seller-client/src/components/StatusStepper/Step.tsx
chriskrogh: yessir 93d4d86249d32b33af3161b8ed37cc3ab96137fb

packages/seller-client/src/components/StatusStepper/Step.tsx
chriskrogh: updated in 3d17e494faa388829277a8ea375b43f3131c3516 to use `theme.palette.action.disabled` so that we can pass it directly as a string on the color prop. But I also updated https://app.clickup.com/t/burady so that we iron out the theme coloring for disabled / inactive colors on the theme

PR #33, opened 2020-09-02
Cache docker images on TravisCI

jtaaa:
### What's Changed

We now cache our docker image builds and layers after our build script passes. We also load the images and layers from the cached `/docker_images` directory. Hopefully our docker images should build faster! We won't have to re-download the NodeJS image each time and hopefully saving layers will speed up our docker image builds.

CU-buqk3f[merged]


jtaaa:
Task linked: [CU-buqk3f Keep Docker build cache between Travis builds if possible](https://app.clickup.com/t/buqk3f)

jtaaa:
Our PR server-build originally took ~6 mins now with saving, caching and loading docker images it takes ~8 minutes 😁 I think it's because the `docker load` and `docker save` commands take long.

It is correctly using the cache though and was able to skip many steps in the `docker build` process

chriskrogh (approved):

PR #34, opened 2020-09-03
Move Tier from UserModel to StoreModel

jtaaa:
### What's Changed

The tier property on the UserModel has been removed and an identical tier property has been added to the StoreModel.

The store's tier will now determine the SYS transaction fee, the number products and images you can upload, and other premium features.

## Fixes

I found two small issues and fixed them:

1. The schema was emitting to the wrong place due to #27 where I made an apollo folder in the utils folder.
2. The server would log the production host when running locally


jtaaa:
Task linked: [CU-buk3dy Move tier to StoreModel from UserModel](https://app.clickup.com/t/buk3dy)

chriskrogh (approved):

PR #49, opened 2020-09-03
Add --frozen-lockfile flag to docker yarn install cmd

jtaaa:
### What's Changed

To ensure that our dependencies are reproducible we want to run `yarn install` with the `--frozen-lockfile` flag.

[`yarn install` docs](https://classic.yarnpkg.com/en/docs/cli/install/)


chriskrogh (approved):
makes sense 2 me!

PR #51, opened 2020-09-04
Calculate SYS fee based on total and store tier

jtaaa:
### What's Changed

We now calculate the Sell Your Stuff transaction fee correctly based on the store's tier and the transaction total.

The current fees I've implemented are:

`Tier.Free`:
- `<= 100`: 2%
- `<= 300`: 3%
- `<= 600`: 3.5%
- `<= 1000`: 4%
- `> 1000 `: 5%

`Tier.Launch`: 2%
`Tier.Growth`: 2%

### Notes

I also remove the `.populate('...')` call on many DB queries in favour of using the GraphQL field resolvers to return these values when requested. This may be less performant than just populating the records (not sure). If this is an issue in the future we can look into optimizations.


jtaaa:
Task linked: [CU-byhwpv Add sysFee calculation](https://app.clickup.com/t/byhwpv)

chriskrogh (commented):
Just curious, but what made you want to swap `populate` with `FieldResolver`? Wouldn't the field resolver run each time anyways? I think it might be less performant that way because instead of going directly to the db, we now have to go through some GQL logic and then go to the db? (assuming that populate and find have the same performance)

jtaaa:
I believe the FieldResolver will only run if the user requests the field in their query. So if the user does not request the field we save a DB call (or populate call). I'm really not sure if the `find` for the FieldResolver is less performant than the `populate` call but I'm a fan of not pre-optimizing (optimize after when a bottleneck arises or it becomes clear an optimization is needed because you may optimize and it has no real-world effect).

I switched to FieldResolvers so that if the user requests a deeply nested property such as

```graphql
products {
  options {
    product {
      options {
        _id
      }
    }
  }
}
```

everything will resolve correctly. Previously with `populate` we could only go so far as the populated fields would allow.

chriskrogh (approved):
LGTM!

PR #52, opened 2020-09-04
[UPSL Client P1] Use UPSL ServiceAreas for address townCity field

jtaaa:
### What's Changed

The type of the `townCity` field on the `AddressModel` is now `ServiceArea` which is an enum of all the valid service areas for the UPSL API 🚚 

I also added an `addresses` array (much like the one on `StoreModel`) to the `UserModel` and a `deliveryAddress` field.


jtaaa:
Task linked: [CU-bykx87 Address town/city field](https://app.clickup.com/t/bykx87)

packages/server/src/utils/upsl/ServiceAreas.ts
```diff
@@ -0,0 +1,230 @@
+import { registerEnumType } from 'type-graphql';
+
+/**
+ * Pulled from the UPSL API
+ * http://domestic.universalpackagesys.com/Test.DSS/v2/serviceareas (requires auth)
+ */
+export enum ServiceArea {
```
chriskrogh, line 7: Will there ever be a need for these values to be consistent with UPSL? The reason I'm asking is because I'm wondering if we'll ever get any errors if this list is inconsistent with what's provided by UPSL. 
I'm a super huge fan of enums, but it might make sense to call the API to get this list and make this a calculated list instead of static if this is being used to communicate with UPSL?

chriskrogh (approved):
LGTM just had a q

packages/server/src/utils/upsl/ServiceAreas.ts
jtaaa: Good point. If `ServiceArea` goes out of date with the UPSL API then our requests will fail.

This enum is used as a GraphQL type. Maybe we can generate this enum dynamically by making a UPSL API request when the server starts.
I think the service areas are unlikely to change often and I like the benefit of defining the enum in the GraphQL schema to avoid having to have a separate query (that generates this freshly) to return these values.

jtaaa: [Ticketed here](https://app.clickup.com/t/byq30h)

**Other ideas:** we can make a cron job that runs periodically and syncs the enum. This might be most appropriate since we expect service areas to be update rarely.

chriskrogh: Ah okay. Yep we can ticket for a future refactor.

PR #53, opened 2020-09-04
[UPSL Client P2] Add UPSL API client

jtaaa:
### What's Changed

We have a UPSL API client! So far, it's capable of:

1. determining if it should request a new access token, refresh the old token or use the current one before shooting a request to the API through `UpslClient.ensureValidTokens()`
2. get estimated shipping cost of an order

I decided to use #ObjectOrientedProgramming **hype** **hype**

It just makes things easier to reason about and ensures we only have one UpslClient instance (with same state - tokens and config) shared throughout the project.


jtaaa:
Task linked: [CU-byhwqp Add UPSL API client](https://app.clickup.com/t/byhwqp)

jtaaa:
Task linked: [CU-bykx87 Address town/city field](https://app.clickup.com/t/bykx87)

packages/server/src/utils/upsl/types.ts
```diff
@@ -0,0 +1,42 @@
...
+    message: string;
+    links: string;
+  };
+};
+
+export type UpslTokensRawData = {
+  access_token: string;
+  refresh_token: string;
+  expires_at: number; // epoch time in seconds
+  token_type: 'Bearer'; // unused
```
chriskrogh, line 15: What does `unused` mean in this context? Do we have to type it if we're not concerned with it?

packages/server/src/utils/upsl/types.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export type UpslTokensData = Omit<UpslTokensRawData, 'expires_at'> & {
+  expires_at: Date;
+};
+
+export type UpslEstimatesData = {
+  shipping_charges: number;
+  // unused
```
chriskrogh, line 24: same here

packages/server/src/utils/upsl/UpslClient.ts
```diff
@@ -0,0 +1,88 @@
...
+  private async ensureValidTokens(): Promise<void> {
+    if (!this.tokens) {
+      return await this.getTokens();
+    }
+    if (this.tokens.expires_at <= new Date()) {
+      return await this.refreshTokens();
+    }
+  }
+
+  async getEstimatedShippingCost(
```
chriskrogh, line 70: Sick!

chriskrogh (approved):
Small question about the `// unused` comments but logic lgtm

packages/server/src/utils/upsl/UpslClient.ts
```diff
@@ -0,0 +1,88 @@
...
+        ship_from: from,
+        ship_to: to,
+        weights,
+      },
+    );
+    return response.data.data.shipping_charges;
+  }
+}
+
+export default new UpslClient();
```
chriskrogh, line 88: Does this guarantee a singleton?
If I import this module in 2 different places throughout the app, am I guaranteed to get back the same object?

packages/server/src/utils/upsl/types.ts
jtaaa: I decided to type the whole response in case in the future we need to use more of the response data. Some fields on the UPSL API, such as this one `token_type` seem fairly useless. As far as I can tell it always returns `Bearer` for our auth method.

I figure it doesn't hurt to include it in the type though since it's a REST API and we dl it as part of the response whether we want to or not.

packages/server/src/utils/upsl/UpslClient.ts
jtaaa: Yep since the constructor (and main class) is never exported. Only a single instance is exported.

packages/server/src/utils/upsl/types.ts
jtaaa: Maybe I should get rid of the `unused` comment altogether. What you think?

I'm also considering that this client can be moved to its own package in the monorepo eventually since the code isn't specific to our app. Possibly also published to the NPM registry 🚀  so anyone else using the UPSL API in NodeJS can use it.

chriskrogh: Ah makes sense. I guess we can keep the whole thing typed for documentation purposes. 
Removing the // unused would be cool too. I think from it's usage we'll be able to tell what's used and what isn't.
But I see why you added it!

packages/server/src/utils/upsl/UpslClient.ts
chriskrogh: Ah makes sense. Sg!

chriskrogh (approved):
LGTM! If we could remove the `// unused`s that would be nice

packages/server/src/utils/upsl/types.ts
jtaaa: sgtm!

PR #56, opened 2020-09-06
Add GCP Cloud Logging through winston

jtaaa:
## What's changed

Whenever we use the winston `Logger` in the server package we will not only log to console and local files but also to Google Cloud Logging. @chriskrogh gimme your GCP email and I can give you access to our GCP project and you can see the logs and errors in the Logging section of the console.

Additionally we now log our internal Apollo server errors and mask them in production. Client facing errors like `UserInputError` and `AuthenticationError` are not logged here as errors. Check notes for more.

## UI

image

## Notes

We want to also log query and reponse data for every Apollo request - not just errors. That's next up for me. Regular `ApolloError`s will be logged as normal responses.


jtaaa:
Task linked: [CU-byqg88 Look into Sentry and other error tracking tools](https://app.clickup.com/t/byqg88)

chriskrogh (approved):
LGTM! Glad to have winston on the team!

PR #58, opened 2020-09-06
Add step and unstep OrderStatus mutations

jtaaa:
## What's changed

Added the `stepOrderStatus` and `unstepOrderStatus` mutations to the `OrderResolver`. I decided to go with step and unstep instead of simply taking an order status with a `updateOrderStatus` mutation to limit the user to what they can do.

We handle updating the order status except for when pickup or custom delivery is enabled. The user only needs to move forward a step or move back a step to undo. They don't need to be able to set an arbitrary order status.


jtaaa:
Task linked: [CU-c0guta Add updateOrderStatus mutation](https://app.clickup.com/t/c0guta)

chriskrogh (approved):
LG! I like that we split these up

PR #59, opened 2020-09-06
Created OrderListItem

chriskrogh:
## What's changed

Created OrderListItem component. 

I wrote this component assuming that the caller has access to an `OrderProductOption`

I don't think the designs on Figma really cover the cases where there were more than one "groups" selected so I improvised. Essentially we try to fit as many on a row as we can and then we wrap to the next line. Not a designer so lmk what u think😅

Also the designs on Figma wanted us to use `body3` for the groups typography, but I don't think `body3` exists so I used `body2`. Let me know if you think the sizing is okay?

## UI
4 groups
Screen Shot 2020-09-06 at 1 05 51 PM
3 groups
Screen Shot 2020-09-06 at 1 06 00 PM
2 groups
Screen Shot 2020-09-06 at 1 06 06 PM
1 groups
Screen Shot 2020-09-06 at 1 06 19 PM
4 groups tablet sized screen
Screen Shot 2020-09-06 at 1 09 54 PM
No Image
Screen Shot 2020-09-06 at 1 06 24 PM

## Notes

Got caught up with webveloper stuff (again) But I finally got all my stuff merged! Hopefully I can finish all this UI stuff by Tues. Look out for Figma comments!


jtaaa:
Task linked: [CU-bcghq3 Add OrderListItem component](https://app.clickup.com/t/bcghq3)

jtaaa (commented):
Yo @chriskrogh the `OrderListItem` component was actually meant to be [this one](https://www.figma.com/file/0YEz4TCVBNlwcg9swNqhTE/V1.0?node-id=783%3A20):
image

Lemme know if that figma link works.


Do you think you can merge this component you made with `ProductListItem`? I originally intended for `ProductListItem` to optionally take a whole bunch of params (status, optionsCount, groupNames **and groups**) and display them if they are passed to the component.

So only if status is provided will it display status, etc. This way it's more flexible to fit many use cases and it's only one component to update if we want to change how it looks.

packages/seller-client/src/components/OrderListItem/OrderListItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+        </Box>
+        {groups && (
+          <Box display="flex" flexWrap="wrap">
+            {groups.map((group) => (
+              <>
+                <Box key={group.name} display="flex" flex={1}>
+                  <Typography variant="body2" noWrap>
+                    {group.name}: {group.option}
+                  </Typography>
+                </Box>
```
jtaaa, line 64: this `Box` only has one child so I don't think we need `display="flex"`
```suggestion
                <Box key={group.name} flex={1}>
                  <Typography variant="body2" noWrap>
                    {group.name}: {group.option}
                  </Typography>
                </Box>
```

packages/seller-client/src/components/OrderListItem/OrderListItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+            height={IMAGE_SIZE}
+          />
+        </Box>
+      )}
+      <Box flex={1} paddingY={1}>
+        <Box display="flex" justifyContent="space-between">
+          <Typography variant="subtitle1">{name}</Typography>
+          <Typography variant="subtitle2">
+            <Price price={price} />
+          </Typography>
```
jtaaa, line 54: Screen Shot 2020-09-06 at 2 52 46 PM

The comment with a "Roboto Regular" after it is the actual variant that we should use. I think the top comment might just be based on the name of the text in Figma (completely unimportant).

Your `body2` for the groups is correct. The first comment is just the name of the text that got auto-incremented from Figma to `body3`

jtaaa (commented):
LGTM it's just this isn't the component I was expecting for `OrderListItem`. Lemme know if you think you can merge this stuff into `ProductListItem`.

PR #60, opened 2020-09-07
Add shipOrderYourself mutation

jtaaa:
## What's changed

Added the `shipOrderYourself` mutation. It requires that:

1. An order with the specified id exists
2. The order is linked to the user's store
3. The order is using Sell Your Stuff delivery
4. The order has status preparing

This makes it so that you can only use the `shipOrderYourself` mutation if the delivery type of the order is `Custom` and the order has not yet been shipped/begun shipping.

## Notes

I made `ensureSysDelivery` and `ensureStatusPreparing` utils that throws since I bet we are going to use em again for other mutations.


jtaaa:
Task linked: [CU-c0gute Add shipOrderYourself mutation](https://app.clickup.com/t/c0gute)

jtaaa:
Task linked: [CU-c0guta Add updateOrderStatus mutation](https://app.clickup.com/t/c0guta)

chriskrogh (approved):

PR #61, opened 2020-09-07
Add cancelOrder mutation

jtaaa:
## What's changed

Added the `cancelOrder` mutation. You can only cancel an order if your store is linked to the order and:

1. The order is preparing if Sell Your Stuff delivery is enabled
2. The order is not complete if Sell Your Stuff delivery is not enabled (Custom delivery or pickup)


jtaaa:
Task linked: [CU-c0gutf Add cancelOrder mutation](https://app.clickup.com/t/c0gutf)

jtaaa:
Task linked: [CU-c0gute Add shipOrderYourself mutation](https://app.clickup.com/t/c0gute)

packages/server/src/models/Order/OrderModel.ts
```diff
@@ -69,6 +70,11 @@ export class Order {
   @DBProperty({ required: true, enum: OrderStatus })
   status!: OrderStatus;
 
+  @GraphQLField()
+  get completed(): boolean {
+    return isCompletedStatus(this.status);
```
chriskrogh, line 75: nit: I think a more ideal name for this function would be `getCompletedStatus` just so that there's a verb in there. I actually expected `isCompletedStatus` to be a boolean variable.

chriskrogh (approved):
Small nit. Otherwise lgtm

packages/server/src/models/Order/OrderModel.ts
jtaaa: to me `getCompletedStatus` might imply that we are getting a completed status from the status we pass in or something like that.

I get your point tho! Whenever I see `is` or `has` I immediately expect a boolean result based on convention as well but I expect it to be either a variable or a function.

I'll refactor to `getIsCompletedStatus`. Should be the best of both worlds!

jtaaa: I'll also ticket a refactor for all our models to use `is` or `has` prefixes where it helps readability

PR #62, opened 2020-09-07
Add custom errors

jtaaa:
## What's changed

We now have two new error types! Introducing the `MetadataError` 🐛  and the `DBError` 🐞 

The `MetadataError` is very generic. It allows us to attach metadata to the error through its trailing rest params. It's a quick and easy way to throw an error with some extra info.

On the other hand we have the `DBError`. It's very specific. By adding constructor overloads each with a different `DBErrorType` for the `type` parameter we can ensure that the error is created with the required information for easy error tracking/handling.

## Notes

I think it's beneficial to use specific error types like this instead of the generic `MetadataError` as much as possible. We don't wanna define too many error classes tho, probably only for large groups of errors like DB errors. Maybe next we'll have `ExternalApiError` or `UpslApiError`


chriskrogh (approved):
!!!

PR #63, opened 2020-09-07
Add getWithRef global static model method

jtaaa:
## What's changed

We now have a `DBSchema` base class for our Typegoose model classes to extend to give the model access to static methods shared between all models. The first global static model method I have added is `getWithRef` which takes a `ref` and a mandatory `source` param (the parent doc - refs must come from somewhere in the DB). It will return the document or throw an error with the source information.

## Notes

I implemented this in at least 3 different ways and made a whole bunch of changes that I had to revert when trying out stuff. I spent **a lot** of time on this but I think it's really useful, for error handling and hopefully for other use cases in the future.

In the end I'm really happy with this solution and it required very minimal changes to existing models.


chriskrogh (approved):
Super smart idea. That's a good way to share logic

PR #64, opened 2020-09-07
Groups on ProductListItem

chriskrogh:
## What's changed
2 things have actually changed with this PR and I definitely want feedback on some of these changes.

1. I adjusted the props of this component because it seemed like we were keeping track of things that the component didn't use.
a. I adjusted the params of the `onClick` prop to not take any params instead of taking an `_id` param. The reason I did this is because we also typed `_id` as a required prop and the only place it was being used was was to call `onClick`. So since the caller will have access to `_id` and the event handler, it can just call the event handler with `_id` itself and pass that function as the `onClick` prop.
b. I removed the `_id` prop because of the above^
c. I didn't see the need for us to take in an array of images as a required prop if we only render the first image if it exists. We now take an optional image prop.
d. We don't use the `_id` field on the `image` prop so I removed it from the props type.

I also want to know what the relationship between `optionCount`, `groupNames` and `groups` is? It seems like groupNames might be redundant if we have groups, but if we don't always have groups, then we'd need `groupNames`.

2. Because of this design:
Screen Shot 2020-09-07 at 11 11 13 AM

We needed to have a way for a product to be displayed with just its name, price and groups.
With the current implementation of `ProductItemList` we couldn't achieve this because we required certain props like `status` and `optionsCount` and displayed them all the time.
To allow just groups, we need to make the other props optional and only render them if they were passed.

Also we rendered "Individual Options" when `groupNames` was falsy. But because sometimes we only want to see just name, price, orders, we needed a way to hide "Individual Options" when `groupNames` was falsy. So to render "Individual Options" we now have to pass an empty array for `groupNames`.

## UI
Everything
Screen Shot 2020-09-07 at 11 52 30 AM

____________
Just Groups
Screen Shot 2020-09-07 at 11 52 36 AM

____________
Status and groupNames
Screen Shot 2020-09-07 at 11 52 42 AM


packages/seller-client/src/components/ProductListItem/ProductListItem.tsx
```diff
@@ -58,22 +59,48 @@ const ProductListItem: React.FC<ProductListItemProps> = ({
...
+          )}
+          {optionsCount && optionsCount > 2 && (
+            <>
+              <Typography variant="caption">
+                <strong>{optionsCount} Options</strong>
+              </Typography>
+              <Spacer width={16} />
+            </>
+          )}
+          {groupNames != null && (
```
jtaaa, line 82: JavaScript with the `!==` vs. `!=`. I think this actually works but it should really be `groupNames !== undefined` or

```suggestion
          {groupNames && (
```

packages/seller-client/src/pages/ProductsPage/ProductsPage.tsx
```diff
@@ -36,6 +36,7 @@ const ProductsPage: React.FC = () => {
           <ProductListItem
             key={product._id}
             {...product}
+            image={product.images?.[0]}
```
jtaaa, line 39: I think `product.images` will always be defined as an array so no need for the optional chaining (`?.`)

if the array is empty it will return undefined anyways for `product.images[0]`
```suggestion
            image={product.images[0]}
```

packages/seller-client/src/components/ProductListItem/ProductListItem.stories.tsx
```diff
@@ -6,25 +6,50 @@ import type { ProductListItemProps } from './ProductListItem';
...
+      <ProductListItem
+        {...REQUIRED_PROPS}
+        status={ProductStatus.Active}
+        groupNames={[]}
+      />
+    </Box>
+  );
+};
+
+export const Playground: React.FC = () => {
```
jtaaa, line 52: Can we add `groups` to the Playground story? I dunno if that would be tough. Not a big deal at all.

jtaaa (changes requested):
Absolutely love all the decisions you made 😍  10/10 for making these decisions with less than perfect knowledge of the use cases and original intention for the component.

I especially love the empty `groupNames` decision 🚀  since it matches the use cases of displaying a product vs. displaying a product option.

Just one problem with the condition on rendering `groupNames` and this is good to go!

jtaaa:
For `optionCount`, `groupNames` and `groups`:

- `optionCount` and `groupNames` only make sense if we are trying to display a `Product`
- `groups` only makes sense if we are trying to display a specific `ProductOption`

These two sets of props likely aren't going to be used together.

packages/seller-client/src/components/ProductListItem/ProductListItem.stories.tsx
```diff
@@ -6,25 +6,50 @@ import type { ProductListItemProps } from './ProductListItem';
...
   title: 'Components/ProductListItem',
   component: ProductListItem,
   decorators: [withKnobs],
 };
 
+const REQUIRED_PROPS: ProductListItemProps = RequiredProductListItemData;
 const BASE_PROPS: ProductListItemProps = ProductListItemData;
 
-export const Base: React.FC<Partial<ProductListItemProps>> = (props) => {
+export const Base: React.FC = () => {
```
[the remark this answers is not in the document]
chriskrogh: Ah interesting to know!

packages/seller-client/src/components/ProductListItem/ProductListItem.tsx
chriskrogh: Yea `!= null` would cover the cases where it's `null` or `undefined` but you're right in the sense that `groupnames` will never be `null`. Updated the check in cff2ec8c7d46a49eff5c08106e62ea67ab17e2fe

packages/seller-client/src/components/ProductListItem/ProductListItem.stories.tsx
chriskrogh: Ah you're right. Updated in 43a0227ca4e4f148467cc9b61425f4da9a4922d9

packages/seller-client/src/pages/ProductsPage/ProductsPage.tsx
chriskrogh: Ah you're right. Updated in 43a0227ca4e4f148467cc9b61425f4da9a4922d9

jtaaa (approved):
Looks good!

PR #71, opened 2020-09-07
Created BuyerDetails component

chriskrogh:
## What's changed
Created BuyerDetails component and smaller sub components that it uses. 
(I'm a fan of splitting components when they get too big)

As of now the only optional props are the `email` and `phoneNumber` because they're also optional on the `UserModel`
`rating` doesn't exist yet, but I still built the stars logic for when it does get built.

For the stars, we can take a rating value `0<= rating <=5` and display the number of stars for that rating.

## UI
With buttons
Screen Shot 2020-09-07 at 6 16 39 PM
___________
Varied stars
Screen-Recording-2020-09-07-at-6


jtaaa:
Task linked: [CU-bcgunb Add BuyerDetails component](https://app.clickup.com/t/bcgunb)

packages/seller-client/src/utils/theme.ts
```diff
@@ -53,6 +53,7 @@ let theme = createMuiTheme({
     MuiButton: {
       root: {
         height: 48,
+        minWidth: 40,
```
chriskrogh, line 56: Had to make this 40 so that the `email` and `phoneNumber` buttons would be circles

packages/seller-client/src/components/BuyerDetails/BuyerDetails.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const VariedRating: React.FC = () => {
+  return (
+    <Box maxWidth={400}>
+      <BuyerDetails
+        {...BASE_PROPS}
+        rating={select(
+          'Rating',
+          [0, 0.33, 1, 1.76, 2, 2.5, 3, 3.01, 4, 4.99, 5],
+          BASE_PROPS.rating,
+        )}
```
jtaaa, line 51: btw this could be a `number` knob

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { Box, Typography, Button } from '@material-ui/core';
+import PhoneIcon from '@material-ui/icons/Phone';
+import MailIcon from '@material-ui/icons/Mail';
```
jtaaa, line 5: Can use a single import

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <Box>
+      <Box display="flex">
+        <Asset
+          size={ASSET_SIZE}
+          initials={`${firstName[0].toLocaleUpperCase()}${lastName[0].toLocaleUpperCase()}`}
+        />
+        <Spacer width={32} />
+        <Box>
+          <Typography>
```
jtaaa, line 54:
```suggestion
          <Typography variant="h6">
```

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
```diff
@@ -0,0 +1,91 @@
...
+    borderRadius: BUTTON_SIZE / 2,
+    backgroundColor: palette.primary.main,
+    color: 'white',
+  },
+}));
+
+type Props = {
+  firstName: string;
+  lastName: string;
+  email?: string;
```
jtaaa, line 30: `email` is actually a required prop on the `UserModel`. In fact the `phoneNumber` prop is also required to be set before the buyer can make a purchase. Both of these props should be required

packages/seller-client/src/components/BuyerDetails/RatingStars.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const RatingStars: React.FC<Props> = ({ rating, size }) => {
+  const style = { width: size, height: size, color: '#FFBA07' };
+  return (
+    <Box display="flex">
+      {STARS.map((_, index) => {
+        const starNumber = index + 1;
+        if (rating >= starNumber) {
+          return <StarIcon style={style} />;
+        } else if (starNumber - rating < 1) {
```
jtaaa, line 22: Having the variables on the same side for each and using greater and greater than makes it easier for me to understand

```suggestion
        } else if (rating > starNumber - 1) {
```

packages/seller-client/src/components/BuyerDetails/RatingStars.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { Box } from '@material-ui/core';
+import StarIcon from '@material-ui/icons/Star';
+import StarHalfIcon from '@material-ui/icons/StarHalf';
+import StarBorderIcon from '@material-ui/icons/StarBorder';
+
+const STARS = new Array(5).fill(1);
```
jtaaa, line 7: Filling these boys with null makes it super clear that the array values aren't necessary.

I think I actually prefer using the `new Array(5).fill(null)` or `range` (from lodash) directly since `STARS` doesn't actually contain any stars. Super nit pick tho. No big deal

```suggestion
const STARS = new Array(5).fill(null);
```

packages/seller-client/src/utils/theme.ts
jtaaa: Why not just add minWidth to the `email` and `phoneNumber` button styles?

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
```diff
@@ -0,0 +1,91 @@
...
+  phoneNumber?: string;
+  rating: number;
+  address: Pick<Address, 'line1' | 'townCity' | 'country'>;
```
jtaaa, line 33: We wanna display `line2` as well if it exists

jtaaa: and `zipCode` and `state`

jtaaa (changes requested):
The component looks really good! Initials are on point 💯 

I added quite a few comments and suggested some changes.

Most importantly, the `email` and `phoneNumber` props should be required. We display this component in the seller-client for sellers to contact buyers if necessary so it must at least show `email` (which is required on the `UserModel`). Also the user must have a `phoneNumber` set in order to make an order so at this point we are guaranteed to have the user's phoneNumber.

jtaaa:
Also looks like our generated `ServiceArea` enum lost its white space somewhere on the journey from the backend 😢 I'll look into that

packages/seller-client/src/components/BuyerDetails/BuyerDetails.stories.tsx
chriskrogh: cool! c6f7f2d58179e6f4a4b3612059170f139c853b17

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
chriskrogh: updated in 10953894bd0fc2ce4482e85f95f148c73cc8fdcd

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
chriskrogh: updated in 5c8429ee1535b9946a5920d4d6f773becd81cfd1

packages/seller-client/src/components/BuyerDetails/RatingStars.tsx
chriskrogh: updated in 8d239d83817e469e5bb88b6324a34735b74cfd55

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
chriskrogh: updated in 661e46a1dd349e400070d099a1744f63098dd39a

packages/seller-client/src/utils/theme.ts
chriskrogh: lmao🤣 idk what had me convinced that this was the only way. updated in cbd7cc9b7602b3d9b3ade18a3b271776b6cab899

packages/seller-client/src/components/BuyerDetails/BuyerDetails.tsx
```diff
@@ -0,0 +1,100 @@
...
+              <br />
+            </>
+          )}
+          {townCity}
+          <br />
+          {(state || zipCode) && (
+            <>
+              {state && state}
+              {state && zipCode && ', '}
+              {zipCode && zipCode}
```
jtaaa, line 79: Don't have to check for existence for these
```suggestion
              {state}
              {state && zipCode && ', '}
              {zipCode}
```

jtaaa (approved):
Looks good!

PR #72, opened 2020-09-08
Make enum string values consistent across packages

jtaaa:
## What's changed

Enum string values are now consistent across `server` and `seller-client`. The enums with values that we need to share have been moved to `shared`.

## Notes

We are using the `enumValues` config option to override the default enum value that `graphql-codegen` generates in the `seller-client`.


jtaaa:
Task linked: [CU-c0pw8v Enums loose whitespace when generated on seller-client](https://app.clickup.com/t/c0pw8v)

chriskrogh (approved):
LGTM!

PR #73, opened 2020-09-09
Add sysDeliveryFee query

jtaaa:
## What's changed

We have a new `sysDeliveryFee` query! It takes order items in the form

```ts
type _ {
  quantity: number;
  items: ObjectId[];
}
```

and returns the **Sell Your Stuff delivery fee**. We use the UPSL API to get the estimated shipping cost.

## Notes

Gotta get on https://app.clickup.com/t/c2pcbw so that this won't fail when a user or store sets a Town/City that isn't a valid ServiceArea.


jtaaa:
Task linked: [CU-c0guu7 Add sysDeliveryFee query](https://app.clickup.com/t/c0guu7)

jtaaa:
Our user facing errors are kinda messy right now. I think the `invalidArgs` prop on the error metadata is very useful but I haven't been using it consistently. I don't think this is a top priority though. We can make changes to the backend if we run into problems down the road.

chriskrogh (approved):
lgtm!

PR #79, opened 2020-09-09
Replace shippingLabelUrl with sysShippingInfo

jtaaa:
## What's changed

Instead of the `shippingLabelUrl` field on the `OrderModel` we now have the `sysShippingDetails` field which stores select information from the UPSL API.

## Notes

We only store the data that would be useful to show to customers in the DB but we will also log the request and response of every request to the UPSL API (coming soon!).


jtaaa:
Task linked: [CU-c0guu7 Add sysDeliveryFee query](https://app.clickup.com/t/c0guu7)

chriskrogh (commented):
This looks very similar to https://github.com/SYSTT/sys-monorepo/pull/73
Maybe the base of this PR is not the right branch?

jtaaa:
It looks like GitHub _finally_ supports automatically changing the base branch when the base branch is merged. But maybe because we squash our commits the changes from the last PR are showing up here even though they are already in master

jtaaa:
Merging `master` into this branch did it so that only the current changes are shown

chriskrogh (approved):

PR #82, opened 2020-09-09
Add trackShipment method to UpslClient

jtaaa:
## What's changed

Another new method on the `UpslClient`. This time it's `trackShipment`.


packages/server/src/utils/upsl/UpslClient.ts
```diff
@@ -134,6 +136,22 @@ class UpslClient {
...
   }
+
+  async trackShipment(
+    trackingNumber: string,
+  ): Promise<UpslTrackShipmentResult> {
+    await this.ensureValidTokens();
+    const response = await this.client.get<UpslResponse<UpslTrackingData>>(
+      `tracking/${trackingNumber}`,
+    );
+    const timeline = response.data.data.timeline.map((timelineItem) => ({
```
chriskrogh, line 147: What happens if the response is unsuccessful? Do we have an error boundary that catches this?

packages/server/src/utils/upsl/types.ts
```diff
@@ -56,12 +56,27 @@ export type UpslLabelData = {
...
+  timestamp: number;
+};
+export type UpslTrackingData = {
+  tracking_number: string;
+  requested_at: number;
+  ship_to: string;
+  status: string;
+  updated_at: number;
+  timeline: UpslTrackingTimelineItem[];
+};
```
chriskrogh, line 71: I think we should reach out to UPSL and ask them how they intend to iterate on their API. I hope they allow backwards compatibility across  API versions.
Just mentioning because our app does have a strong type dependency on their api responses (which is okay) but if they change something I don't want us to have to deal with any crashes

chriskrogh (approved):
Loving the smaller chunk sizes bro. < on the 75 diffs😍

packages/server/src/utils/upsl/UpslClient.ts
jtaaa: Currently no. If the `UpslClient` fails or throws an error for any reason we return an `INTERNAL_SERVER_ERROR`. We may want to tweak this to throw a specific, possibly client-facing `UpslError` so we can display an error to the user indicating that UPSL's API failed unexpectedly (please try again or try again later).

I plan on making the `Axios` client in that the `UpslClient` uses log all of its requests and responses so when a request fails we will know and have all the inputs and outputs to figure out what happened.

packages/server/src/utils/upsl/types.ts
jtaaa: Yeah as far as I can tell the API isn't actively developed so I don't expect major changes but I'll get the developers contact from @jarryon and send an email just asking him to keep us in mind if there are changes.

PR #83, opened 2020-09-09
Add getFromUserInputId method to DBSchema

jtaaa:
## What's changed

Added the `getFromUserInputId` method to the DBSchema that our model classes extend. We can use this in the backend to retrieve a doc by `_id` or throw a `UserInputError` if the document does not exists. The point is that when handling user input `id`s we always want to throw a `UserInputError` if the specified doc does not exist.


chriskrogh (approved):
Very smart idea

PR #84, opened 2020-09-09
Add folders and makePublic option to uploadFile

jtaaa:
## What's changed

The `uploadFile` util is now even more useful. We can specify a folder to upload the file to instead of the bucket as previously since we are using a single bucket for our domain `sellyourstuff.online`.

We now have the option `makePublic` which will make the file publicly accessible to the internet. Previously it would automatically make the file public.

## Notes

We don't want to make our PDF shipping label files public because we want to restrict access to only the store owners. After I add the `generateShippingLabel` mutation to the `OrderResolver` I'll make a regular Express endpoint that streams the file to the user if they are the owner of the store related to the order.


chriskrogh (approved):
1. Since we can now specify folders to store assets in gc buckets, what approach are we taking when storing assets? Will every store have a folder? What about products without stores? Just curious.
2. If you didn't already can you follow up with a ticket for the regexp filter that you mentioned in the notes😅 I know you won't let it get lost but I think we should keep track of as much work as possible through CU no matter how small

jtaaa:
2. Yessir forgot to link this https://app.clickup.com/t/c4mkh8

1. I was planning on storing all product images in one folder and all shipping labels in another.

The folder stuff is really just for convenience of browsing and accessing the files I believe. The folders are really just filenames with `folder/...` indicating a folder.

We can make folders for each store and have stuff sorted really well but I don't see any big advantage. It will be just as hard if not harder to keep track of the files in our apps.

PR #85, opened 2020-09-09
Add generateShippingLabel mutation

jtaaa:
This is the big guy!

## What's changed

We now have the `generateShippingLabel` mutation which does 3 things (all through the UPSL API):

1. Creates a UPSL shipment
2. Gets the shipping label for the shipment
3. Tracks the shipment to populate the initial timeline

It will also upload the shipping label pdf file to the GCS in `ShippingLabels` folder.

## Notes

The client will need to run this mutation and then open a new tab on a backend endpoint (not built yet 🚧 ) to display the PDF.

This takes kinda long since we have to round-trip to UPSL's API 3-4 times. It also stacks up our 100 requests per minute UPSL rate limit. We can potentially remove the initial timeline population but let's see if it's an issue first.

In Postman the request takes between 2 and 4.5 seconds to complete (depending on whether we have valid tokens) . Not too sure how we can cut that down. We'll need to display some kind of loading indication on the frontend.


chriskrogh (approved):
lgtm!
Yea what would be ideal is if the server could send little responses when each one of those reqs complete so that we could use those to update the UI in like a step loading (bar?) component. I think you might have to use sockets for that? Not sure. It would be cool to know how other people do it.

I'm fine with having a longer loading time, we just need a way for the UI to reflect what's going on

jtaaa:
Truee that's a possibility.

We also have the option of splitting this up into 3 requests and having the client worry about it.
Though the advantage of making one request on the server side is we can undo the previous requests (cancel shipment through UPSL API) if a subsequent request fails. Although we don't have that set up yet 😢

jtaaa:
I'm pretty sure we can achieve the _socket like_ progress indication through GraphQL subscriptions. Haven't explored those yet tho

PR #88, opened 2020-09-10
Shared Components

chriskrogh:
This PR looks pretty huge, but that's because `packages/shared/graphql.schema.json` is like ~6700 +s and `packages/shared/generated/graphql.tsx` was 800 +s.
I recommend viewing changes by commits.

## What's changed
1. I moved these to `shared/src/components`
Screen Shot 2020-09-10 at 11 56 05 AM

2. We needed to update dependencies for shared just to include `react`, `@material-ui/core` and `@material-ui/lab` because we declare types and modules in `theme` that required those libs.

3. We needed codegen to run for the shared components. This means that shared components have a dependency on server. This should be okay for now since `server` only depends on the other modules of shared and not shared components.

4. Needed to update the imports of the components in sc that call the 4 shared components.


jtaaa:

jtaaa:
Yo @chriskrogh it's kinda tricky to export subdirectories. NodeJS 14 adds the `exports` field in the `package.json` but webpack does not support it yet. It will when webpack v5 is released 🥳 On the other hand CRA doesn't support it yet and we can't assume when webpack 5 comes out CRA will support it immediately.

Right now I'm exporting everything from the index and it seems to work fine. It's kinda lame tho and leads to clutter when importing. I also had to get rid of absolute imports in the shared package since we cannot use our `NODE_PATH`. We might have to look into building with Webpack.

I'm pretty sure there will be an issue with a circular dependency on the enums in shared. The server uses them to generate the schema and then shared generates it's own

jtaaa:
I got rid of the `generated/graphql` and re-added the `enums` export so that the server can import from there safely.

This should fix the circular dependency issue?

We should actually have all our generated files ignored in git.

jtaaa:
Hooray! It builds

chriskrogh:
Big up!!! Let's goooo

jtaaa (approved):
letssgoo

chriskrogh:
lgtm !

PR #90, opened 2020-09-12
Throw UserInputError for incorrect password/email

jtaaa:
## What's changed

In production internal server errors are masked so we don't send the error messages from external packages. We have to handle errors from these packages and throw appropriate `ApolloError`s.


jtaaa:
Task linked: [CU-c8k2we SignIn mutation throws Internal Server Error for bad creds](https://app.clickup.com/t/c8k2we)

chriskrogh (approved):

PR #96, opened 2020-09-12
Add server endpoint to view shipping labels

jtaaa:
## What's changed

We now have a server endpoint that takes the `orderId` as a URL param and serves the shipping label PDF file if the user is **authenticated** and **owns the store that the order is related to**.

## UI

Screen Shot 2020-09-12 at 7 02 59 AM

## Notes

I also fixed an issue with Typegoose where it was failing to implicitly determine the type of an input. It was also complaining about the `store` field ref on `ProductModel`. I'm very confused by this as it is only happening sometimes. @pratyushpal got these errors initially too but a clean clone cleared it up for him.


jtaaa:
Task linked: [CU-c4mkh8 Add server endpoint to view shipping labels](https://app.clickup.com/t/c4mkh8)

chriskrogh (approved):

PR #121, opened 2020-09-14
Replace Product image mutations with ImageResolver

jtaaa:
## What's changed

Instead of having `addProductImages` and `removeProductImages` mutations we now have a generic `uploadImages` mutation. The `NewProductInput` type that the `addProduct` mutation takes already includes an `images` array of `ObjectId`s which we must populate after uploading the images.

## Notes

We may need to pay more special attention to images when we implement store logos and user images. These images (or at least the store logo) _may_ not be required to be square. We also may not need a thumbnail image for them.
I think it's best to cross that bridge when we get there.


jtaaa:

chriskrogh (approved):
LGTM

PR #124, opened 2020-09-14
Setup for uploading images

jtaaa:
## What's changed

Added the `useUploadImages` hook for uploading images without touching the `use...Mutation` hook. It throws to ensure that the `uploadImages` async function always returns an `Image[]`.

We can use Error Boundaries to catch these errors, potentially limiting the error to only the `ImagesUploader` component (while the rest of the UI will still work? _maybe hopefully_).


jtaaa:

chriskrogh (approved):

PR #125, opened 2020-09-14
Add initial categories query to ProductResolver

jtaaa:
## What's changed

We have a shared `ProductCategory` enum. We also have `getStringEnumKeys` and `getStringEnumValues` util functions.

I've updated the type of the category field on the `ProductModel` to be a `ProductCategory`.

## Notes

In the future we may want to do more with categories like storing them in the DB and allowing admins to dynamically update them. For now having them static is fine.


jtaaa:

chriskrogh (commented):
Why do we need this? If this is just a static array, then can we just keep this in shared / client depending on where it's needed? How come we have to fetch it?

jtaaa:
You're right! It really doesn't need to be fetched at all if it's static. I'll remove this query and move the categories to `shared`

In the future we may move it back to a query and add the categories to a DB table to make them dynamic.

chriskrogh (approved):
Is there a reason why we'd want to store these in the db in the future?

jtaaa:
Not definitely but we may want to allow admins to update/add categories dynamically instead of releasing a new build. Maybe we'll find that this isn't necessary at all.

PR #126, opened 2020-09-14
Add useTags hook that handles updating cache

jtaaa:
## What's changed

Added the custom `useTags` hook that handles updating the cache for us.
Also made the backend use Title Case for tag values.

## Notes

Right now anyone can add tags and they are publicly displayed to everyone else. This allows stores to select tags that are already in use which will increase the likelihood that their products are found through the tags. However it also allows malicious users to add anything they want as a tag and have other users see it.

We will have to deal with this somehow. Maybe adding a `public` field to tags which can be edited by an admin which will allow tags to be viewable by everyone (default is not viewable to others). Or maybe just allowing admins to prune the list of tags.


jtaaa:

chriskrogh (approved):

PR #132, opened 2020-09-17
Add CopyableTextField component

pratyushpal:
## What's changed

Added a CopyableTextField which will allow users to copy their text/numerical input to their clipboard with the click of a button! 

## UI

AddCopyableTextFieldComponent

## Notes
This is a WIP branch. I'm not sure what needs to be done about the styling here. Any tips/recommendations are welcome! 

This took me a while to finish but I feel a lot more confident about React and using Hooks moving forward.


jtaaa:
Love the GIF bro 😍 will check this out tomorrow!

packages/seller-client/src/components/CopyableTextField/CopyableTextField.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+import CopyableTextField from './CopyableTextField';
+
+
+export default {
+  title: 'Components/CopyableTextField',
+  component: CopyableTextField,
+};
+
+
+export const DummyField: React.FC = () => {
```
jtaaa, line 12:
```suggestion
export const Base: React.FC = () => {
```

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return(
+    <Box>
+      <TextField 
+      inputRef={textAreaRef}
+      > </TextField>
+      {
+       /* Logical shortcut for only displaying the 
+          button if the copy command exists */
+       document.queryCommandSupported('copy') &&
```
jtaaa, line 55: We can remove this comment by using a self-documenting variable name like `const isCopySupported = document.queryCommandSupported('copy') ;`

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <div>
+          <Button onClick={() => handleCopy(textAreaRef)}>
+          {isCopied ? <Check/> : <Assignment/>}
+          </Button> 
+        </div>
```
jtaaa, line 60: This should be included in the [`endAdornment`](https://material-ui.com/components/text-fields/#input-adornments) on the `MuiTextField`

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const text = textField.current.value; 
+    if (typeof text === "string" || typeof text == "number") {
+      setCopied(true);
+      textField.current.select();
+      document.execCommand('copy');
+    } else {
+      setCopied(false);
+      console.error(
+        `Cannot copy typeof ${typeof text} to clipboard, must be a string or number.`
+      );
```
jtaaa, line 21: Lets go ahead and **throw** an error here since the ref we send to `handleCopy` should always have a `number` or `string` value. This should help us catch incorrect usage immediately.

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -0,0 +1,69 @@
...
+import {TextField,Button,Box,} from '@material-ui/core';
+import {Assignment, Check} from '@material-ui/icons'
+
+
+export function useCopyToClipboard(resetInterval: number) {
+
+  const [isCopied, setCopied] = React.useState(false);
+
+
+  const handleCopy = React.useCallback((textField: any) => {
```
jtaaa, line 11: We need to type the `textField` param. Maybe choose a better name since this hook is re-useable 💯  on any `Ref<HTMLElement>`. Maybe just `ref` or `textElementRef`.

jtaaa (changes requested):
Yoo Pratyush this looks really good buddy. Love the hook so that we can reuse the functionality elsewhere!

There are a few changes necessary before we can get this merged in. Firstly is you need to get your editor set up with `eslint` or run `yarn lint` from the `packages/seller-client` dir. This will lint the code and make sure the formatting is consistent.

It will also show you all the `eslint` warnings and errors that your current branch has such as `typescript-eslint/no-explicit-any` which would have warned you against using the `any` type.

Next, the hook is important enough that it can be moved to its own file, possibly even moved to the utils folder. I don't consider it a part of the `CopyableTextField` component but rather an independent piece of re-useable code that is used by the `CopyableTextField`.

Finally, the `CopyableTextField` as it is right now can allow the user to copy things that they input but does not allow us to specify what is in the `MuiTextField` or stop them from overwriting it. I guess the spec could have been a bit clearer but we 100% need to be able to specific the text that will be copied when the copy icon is clicked and also optionally stop the user from being able to edit that text with a `disabled` prop.

Ideally we can take all the props that `MuiTextField` takes and spread them onto the `MuiTextField` with the extra `inputRef` and `endAdornment` props.

I've also added a bunch of comments 💬  Check them out.

Let me know if this makes sense! Thanks for help bro this actually looks good despite taking so long.

jtaaa:
The [build](https://travis-ci.com/github/SYSTT/sys-monorepo/jobs/386021113) is also failing on Travis, seems like `handleCopy` may be typed incorrectly. The build needs to pass before we can merge. It should build in Travis if it builds locally.

packages/seller-client/src/components/CopyableTextField/CopyableTextField.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Box maxWidth={600}>
+      <CopyableTextField
+        variant="outlined"
+        label="Order Support ID"
+        defaultValue="5f1f945f7b1edd95cac6ea66"
+      />
+    </Box>
+  );
+};
```
jtaaa, line 35: Duplicated above

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useRef } from 'react';
+import { TextField, IconButton } from '@material-ui/core';
+import { Assignment, Check } from '@material-ui/icons';
+
+import type { TextFieldProps } from '@material-ui/core';
+import useCopyToClipboard from '../../utils/hooks/CopyToClipboard';
```
jtaaa, line 6: We can use absolute imports in the `seller-client` app like so
```suggestion
import useCopyToClipboard from 'utils/hooks/CopyToClipboard';
```

packages/seller-client/src/components/CopyableTextField/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './CopyableTextField';
```
jtaaa, line 1: You're missing a line ending here. Make sure `eslint` is auto-formatting your work on save
```suggestion
export { default } from './CopyableTextField';

```

packages/seller-client/src/utils/hooks/CopyToClipboard.ts
```diff
@@ -0,0 +1,36 @@
...
+          'The type of text must be a string or number. Cannot copy otherwise.',
+        );
+      }
+    },
+    [],
+  );
+
+  React.useEffect(() => {
+    let timeout: NodeJS.Timeout;
+    if (isCopied && resetInterval) {
```
jtaaa, line 27: This is cool that if `resetInterval` is 0 we don't reset the `isCopied` field. Was this intended behaviour? Maybe we should document this in a comment!

packages/shared/src/generated/graphql.tsx
```diff
@@ -0,0 +1,573 @@
...
+import { ProductStatus } from '../enums';
+import { OrderStatus } from '../enums';
+export type Maybe<T> = T | null;
+export type Exact<T extends { [key: string]: unknown }> = {
+  [K in keyof T]: T[K];
+};
+/** All built-in and custom scalars, mapped to their actual values */
+export type Scalars = {
+  ID: string;
+  String: string;
```
jtaaa, line 13: We aren't tracking generated files anymore since #135 so you can use `git rm --cache -r packages/shared/src/generated` to get rid of this guy in your branch

packages/seller-client/src/utils/hooks/CopyToClipboard.ts
```diff
@@ -0,0 +1,36 @@
...
+    (textField: RefObject<HTMLInputElement>) => {
+      const text = textField.current?.value;
+      if (typeof text === 'string' || typeof text == 'number') {
+        setCopied(true);
+        textField.current?.select();
+        document.execCommand('copy');
+      } else {
+        setCopied(false);
+        throw TypeError(
+          'The type of text must be a string or number. Cannot copy otherwise.',
```
jtaaa, line 18: The caller actually doesn't know what `text` is so she'd be super surprised to see an error about the type of `text`.

A better error might be something like: `The textField ref provided is either undefined or does not contain a string or number value`

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
jtaaa: @pratyushpal how about renaming `textField` to `inputRef` or `ref` so that it's clear it is a `RefObject<HTMLInput>`

packages/seller-client/src/utils/hooks/CopyToClipboard.ts
pratyushpal: yup! that was the intended behaviour

pratyushpal: Added the comment! Though it doesn't show up here as this being outdated.

packages/seller-client/src/utils/hooks/CopyToClipboard.ts
```diff
@@ -0,0 +1,37 @@
+import React, { RefObject } from 'react';
+
+export default function useCopyToClipboard(
+  resetInterval: number,
+): [boolean, (ref: RefObject<HTMLInputElement>) => void] {
+  const [isCopied, setCopied] = React.useState(false);
+
+  const handleCopy = React.useCallback(
+    (textField: RefObject<HTMLInputElement>) => {
```
jtaaa, line 9: I think my previous comment about this got lost. To me `textField` implies `MuiTextField` and it doesn't imply that it's a ref. Maybe `inputRef` is a better name.

packages/seller-client/src/utils/hooks/CopyToClipboard.ts
```diff
@@ -0,0 +1,37 @@
...
+        );
+      }
+    },
+    [],
+  );
+
+  React.useEffect(() => {
+    let timeout: NodeJS.Timeout;
+    if (isCopied && resetInterval) {
+      // If the resetInterval is 0 we do not reset the isCopied field
```
jtaaa, line 28: If you put this comment above the function definition as a [JSDoc comment](https://jsdoc.app/index.html) we will get this information on hover of the function name in VSCode. It's super easy to add as well. Typing `/**` above the function in vscode should add all the boilerplate stuff for you.

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Assignment, Check } from '@material-ui/icons';
+
+import type { TextFieldProps } from '@material-ui/core';
+import useCopyToClipboard from 'utils/hooks/CopyToClipboard';
+
+const CopyableTextField: React.FC<Omit<TextFieldProps, 'inputRef'>> = (
+  textFieldProps,
+) => {
+  const [isCopied, handleCopy] = useCopyToClipboard(3000);
+  const textAreaRef = useRef<HTMLInputElement>(null);
```
jtaaa, line 12: `inputRef` instead? this isn't a `textArea`

jtaaa (approved):
Yoo Pratyush looking better and better!

I've added a bunch more comments. Some things are me being a bit picky with function param names, etc. Some are small improvements that aren't strictly necessary like the JSDoc stuff.

I've approved so we can get this in but please feel free to implement some of the improvements I suggest before you merge.

On top of the comments, the `CopyToClipboard` file should be renamed to `useCopyToClipboard` to match all the other hooks in the `utils/hooks` folder.

packages/seller-client/src/utils/hooks/CopyToClipboard.ts
jtaaa: Don't forget to update the error message too!

packages/seller-client/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const isCopySupported = document.queryCommandSupported('copy');
```
jtaaa, line 14: This could be returned by the `useCopyToClipboard` hook

PR #133, opened 2020-09-17
Shared Theme and Storybooks

chriskrogh:
// `yarn.lock diffs r big`

## What's changed

Our theme has been completely moved to the shared package. This is because we have some components in there that need the theme. I'm also thinking that we'll also most likely be sharing this theme with `marketplace`?

I also set up `storybook` in shared with the same plugins that we use in sc.

## Notes
Next I'll be adding the shared component `OrderDetails` that all of this set up is ACTUALLY for😂


jtaaa:

packages/shared/.storybook/preview.js
```diff
@@ -0,0 +1,3 @@
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
```
jtaaa, line 2: cool!

jtaaa (commented):
LGTM just some newlines at EOF are missing. Our eslint setup probably isn't linting JS

chriskrogh:
updated eof spacing in 0462245105426281beaa770647fcbc8e73a5e242

jtaaa:
darnit it didn't close the PR

PR #134, opened 2020-09-17
Shared eslint setup tweaks and generated file changes

jtaaa:
## What's changed

Our eslint config for the `shared` package now uses `plugin:@typescript-eslint/recommended`, `plugin:jsx-a11y/recommended`, `prettier/react` in addition to others it used before.

## Notes

Our linting should be consistent across the `seller-client` and `shared` now which begs the question why we don't have a shared eslint config to extend from for these two packages. [chore - Add shared eslint config for client apps](https://app.clickup.com/t/ceteuc)

I also included some generated files that were not included in the previous PR. Ideally we shouldn't be tracking these files. [chore - Stop tracking generated files](https://app.clickup.com/t/cetfrc)


chriskrogh (approved):
LGTM
Those are 2 good recommendations. We can even make the eslint config global across all packages in the monorepo. I can't see a case where we'd want a different eslint config depending on the package? We can discuss later when someone picks up the chore

PR #137, opened 2020-09-21
ProductListItem variants

chriskrogh:
## What's changed

Split the popular `ProductListItem` component into type safe variants:
1. Product variant
2. Order variant
3. Cart variant

We decided to do this to remove the redundancy between the `groups` and `groupNames` props. We also wanted to allow and `order` to render a `ProductListItem` using a `ProductOption` instead of a product so that there's 1 less populate that we have to do.

This PR adds the variants and integrates them into the sc package.

## UI

UI hasn't changed. I added a Cart variant with 2 buttons, but I didn't style the buttons (we can do that when we get to marketplace)

Product variant 
Screen Shot 2020-09-21 at 9 29 32 AM

Order variant
Screen Shot 2020-09-21 at 9 29 59 AM

Cart variant
Screen Shot 2020-09-21 at 9 31 07 AM

## Notes

I love that we have storybooks for shared components!


jtaaa:
Task linked: [CU-cgnj87 Refactor ProductListItem to contain multiple variants](https://app.clickup.com/t/cgnj87)

chriskrogh:
@jtaaa right now, the cart variant and the order variant are very similar, the cart variant just has extra buttons. How would you feel about removing the cart variant and then just adding the buttons below when we eventually do marketplace?

packages/seller-client/src/pages/ProductsPage/ProductsPage.tsx
```diff
@@ -45,6 +46,7 @@ const ProductsPage: React.FC = () => {
             image={product.images[0]}
             optionsCount={product.options.length}
             onClick={() => openProduct({ id: product._id })}
+            variant={ProductListItemVariants.PRODUCT}
```
jtaaa, line 49: for convenience I prefer just using `"cart"`, `"order"` and `"product"` for types for convenience
so we can just type `variant="cart"`

TS's literal types are pretty cool and [template literal types in 4.1](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/) are even cooler

Any benefits to using enums?

packages/shared-components/src/components/ProductListItem/OrderVariant.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+
+import { ProductOptionGroup } from '../../generated/graphql';
+import Price from '../Price';
+import Spacer from '../Spacer';
+import { container } from './styles';
+
+const useStyles = makeStyles(({ spacing }) => ({
+  container: container(spacing),
+}));
```
jtaaa, line 12: Love this 😍 
if we wanted we could take it a step further and make a shared `useProductListItemStyles`

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
```diff
@@ -0,0 +1,81 @@
...
+          />
+        </Box>
+      )}
+      <Box flex={1} paddingY={1}>
+        <Box display="flex" justifyContent="space-between">
+          <Typography variant="subtitle1">{name}</Typography>
+          <Typography variant="subtitle1">
+            <Price price={price} />
+          </Typography>
+        </Box>
```
jtaaa, line 55: This stuff is repeated throughout all the variants. A little bit 💧 (not DRY)

packages/shared-components/src/components/ProductListItem/mocks.ts
```diff
@@ -1,6 +1,11 @@
+/* eslint-disable no-console */
```
jtaaa, line 1:
```suggestion
```

jtaaa (approved):
The problem with adding them below on the Marketplace is that they'll align below the image.

If we wanna make the variant more general we can add an array of buttons specified in props so they can do custom actions ¡¡ I'm fine with this tho!  If we see that it'll be useful later we can refactor

packages/seller-client/src/pages/ProductsPage/ProductsPage.tsx
jtaaa: I don't think the variant ever needs to be updated dynamically

packages/shared-components/src/components/ProductListItem/OrderVariant.tsx
chriskrogh: I was thinking about that, but then if a variant wanted to add its own classname style for another element, it couldn't. I think extracting the whole hook wouldn't allow much flexibility there.

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
chriskrogh: Yea I feel you. It doesn't feel cool now, but I think extracting those small components might be a bit overkill? We're already quite low down in the tree and I don't think this is gonna change too much?

packages/seller-client/src/pages/ProductsPage/ProductsPage.tsx
chriskrogh: Yessir! That's a pretty good idea!

chriskrogh: swapped in 7b2568d2537c0f9a8df17117f1149281042a0260

chriskrogh:
Merging for now but feel free to continue discussion here

packages/shared-components/src/components/ProductListItem/OrderVariant.tsx
jtaaa: I was thinking we could still have the regular `const useStyles = makeStyles(...);` but also import `useProductListItemStyles` and use it in the component alongside `useStyles`

so you'd have `classes` and `sharedClasses`/`productListItemClasses`. so it would be v flexible

def no need to change this tho!

jtaaa: just what I was thinking!

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
jtaaa: I think I would almost always extract because it would be 3x the work, in this case, later on if we needed to update the image (maybe to have them autoscroll through multiple images or something cool like that - even just to have a border)

more realistically it would just be the work of extracting later down the line as a chore before making the change. It would be cool to use this as a test to learn if extracting would be useful later on or if it's too granular

jtaaa: another option would be to flip the layout so that there is one base and 3 variants which each render the base with its props defining the variant-specific content

like
```ts
const ProductVariant = ({ ...props }) => {
  return <ProductListItemBase variantContent={<Box> ... </Box>} />;
};

const ProductListItemBase = ({ variantContent, ...props }) => {
  return (
    <Box>
      // ... all the shared content
      {variantContent}
    </Box>
  );
};
```

jtaaa:
unlucky with this build bruv @chriskrogh 🤣

chriskrogh:
> unlucky with this build bruv @chriskrogh 🤣
> 
> https://travis-ci.com/github/SYSTT/sys-monorepo/jobs/390213993

lmao who would have thought

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
chriskrogh: Yep we can do that! I'm down for it

packages/shared-components/src/components/ProductListItem/OrderVariant.tsx
chriskrogh: I actually prefer that method! Just because it would be more clear where the shared classes are coming from

chriskrogh: This got overwritten by the new base component, but I think we should follow that convention for shared styles!
Pretty clean soln

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
chriskrogh: Created base in 30edd2590ccf306e6883dff2fbbf1bf6b5ee6832

chriskrogh: Just for some context as to why I was hesitant to make a base, the concept of making wrapper components and then passing children is called abstraction and Antonio hated abstraction. He said that it used to reduce flexibility. But imo I think abstraction is pretty cool and we can use it for stuff like this. We're pretty far down the tree so I don't think we reduce flexibility by using it here

jtaaa: That's good to keep in mind. It does reduce the flexibility of each individual variant since they can only specify variant specific content in one part of the component - wherever `children` is rendered. For example we can't have a variant-specific image or title. Our designs don't specify that we will ever need to tho!

I bet Antonio has some good reasons for hating this kinda of abstraction but I'm not a fan of getting so set against something that we can't consider the pattern's usefulness in a specific scenario.

PR #147, opened 2020-09-23
Created OrderDetails component

chriskrogh:
## What's changed

First shared component from birth🥳
I also adjusted the config for codegen so that `.graphql` files are interpreted.

## UI

Screen Shot 2020-09-22 at 10 36 22 PM


jtaaa:

jtaaa:
Task linked: [CU-cgnj87 Refactor ProductListItem to contain multiple variants](https://app.clickup.com/t/cgnj87)

packages/shared-components/src/components/OrderDetails/OrderDetails.tsx
```diff
@@ -0,0 +1,39 @@
+import type { OrderItemOptionFragment } from '../../generated/graphql';
+
+import React from 'react';
+import ProductListItem from '../ProductListItem';
+import Spacer from '../Spacer';
+import TotalBreakdown from '../TotalBreakdown';
+import { VARIANTS } from '../ProductListItem/types';
```
jtaaa, line 7: darn I wish we could do `components/ProductListItem`

jtaaa (approved):
letssgoo LRGTM

packages/shared-components/src/components/OrderDetails/OrderDetails.tsx
chriskrogh: Bro I feel you on this 100%. This is an issue that I have with mern as well. 
Something to look into is  paths in `tsconfig.json` https://stackoverflow.com/questions/43281741/how-to-use-paths-in-tsconfig-json . They meight be more reliable than `baseUrl` ? Idk haven't tried them yet.

jtaaa: `paths` would have the same issue as `baseUrl` unfortunately!
we need Webpack or another bundler to link the JS imports properly. `tsc` won't link them for us.

We use a hacky `NODE_PATH` environmental variable for the server. It tells node where to look for modules 😅  and then resolves our absolute imports as modules. so pretty hacky!

And it doesn't work when the package is meant to be imported by another package - can't set the env var since the package isn't running the script

so  `shared` and `shared-client` can't use the `NODE_PATH` hack

I'll add task for Webpack and getting rid of the `NODE_PATH` hack

PR #157, opened 2020-09-27
Add usdPrice field to ProductForm

jtaaa:
## What's changed

We now have the `usdPrice` field in the `ProductForm`.

This is one of the first `Tier` based features! If the store's tier is not `Growth` this field will be locked and set automatically based on the regular TTD `price` field.

## UI

`Tier: Free`
Screen Recording 2020-09-27 at 6 51 28 PM

`Tier: Growth`
Screen Recording 2020-09-27 at 6 53 47 PM

## Notes

We should really have `Tier` as a numerical enum so that we can do `Tier >= Growth`. [CU-cwr4uw](https://app.clickup.com/t/cwr4uw)

I'm also using a constant `USD_TTD_RATIO` in `getDefaultUsdPrice` in the `shared` package.
This feels a little bit like a red flag 🚩
What do you think? I'm thinking that we can ignore the set `usdPrice` for `Tier < Growth` in the marketplace and always calculate the `usdPrice` on-the-fly. This would allow us to update the ratio and not have to update all the `usdPrice`s on the products.

With the current setup if/when the TTD-USD conversion ratio changes we will need to roll out a new build but we save on a lot of the complexities of fetching the actual current conversion ratio


jtaaa:
Task linked: [CU-cwqzj1 Add usdPricing section](https://app.clickup.com/t/cwqzj1)

jtaaa:
Task linked: [CU-cwr4uw Convert Tier to numerical enum](https://app.clickup.com/t/cwr4uw)

chriskrogh (approved):
A possible solution would be to write a service that queries a exchange rate API for the exchange rate once a day and stores it in a db collection. Then we can query our server for the exchange rate and have it be consistent across sc and mp. Once the order is placed, we can save the current rate onto the order for history preservation.

Lmk what you think and we can ticket this

jtaaa:
That sounds great to me!

Do you think this should be a requirement for v1 or can we survive w/o? Big up **Vings** btw

PR #158, opened 2020-09-27
Convert Tier enum from string enum to numeric enum

jtaaa:
## What's changed

`Tier` doesn't need to be a string enum! By using a numeric enum instead we can compare tiers to restrict or enable features for ranges of `Tier`s easily.

## Notes

~This makes it a little more difficult to display the name of the tier to the user since that information is not encoded in the values of the enum. Instead of just using the enum value we may have to make some kind of shared helper to return the tier's name.~

~Alternatively we could've made a shared helper(s) to compare the enums like `isAtLeastTier(Tier.Growth)` but I think this way is simpler and we rarely, if ever, use the tier name.~

yoo @chriskrogh I messed up pretty royally by assuming that Apollo-Client was handling serializing and deserializing our enums. In reality graphql-codegen was just letting us override the types 🥴  I've found the docs to be pretty confusing probably because it is used on client-side and server-side. [More info here](https://app.clickup.com/t/cwtc5y)

So basically all our GraphQL generated enums are broken. Luckily I think you haven't requested enums from the server in your orders stuff as yet? Then the fix should be as simple as converting to the correct enum type when passing props to components when you implement the fetching from the server. + updating imports.


jtaaa:
Task linked: [CU-cwr4uw Convert Tier to numerical enum](https://app.clickup.com/t/cwr4uw)

jtaaa:
Task linked: [CU-cwqzj1 Add usdPricing section](https://app.clickup.com/t/cwqzj1)

jtaaa:
Oh actually the [reverse mapping](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings) that TS includes in numeric enums might make it really easy to display the tier name!

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
```diff
@@ -29,7 +28,7 @@ const ProductSubPage: React.FC<ProductSubPageProps> = ({ close, id }) => {
 
   // tier
   const { data: tierData } = useTierQuery();
-  const tier = tierData?.store?.tier ?? Tier.Free;
+  const gqlTier = tierData?.store?.tier;
```
jtaaa, line 31: Calling this `gqlTier` to specify that the type comes from the "value-less" enum type in the GraphQL schema.
Apollo-client will return enum values as the names of the enum members and GraphQL codegen will type this correctly if we don't override the `enumValues` field.

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
```diff
@@ -96,7 +95,7 @@ const ProductSubPage: React.FC<ProductSubPageProps> = ({ close, id }) => {
           product={product}
           tags={tags}
           addTag={addTag}
-          tier={tier}
+          tier={Tier[gqlTier ?? 'Free']}
```
jtaaa, line 98: Now we can get the proper internal values of the enum by indexing the enum imported from `shared` with the enum value that `apollo-client` returns. This works since `apollo-client` returns the enum members name as the value.

As a requirement our shared enum member names must match the GraphQL enum member names. Since we are using the shared enums in `server` to create the schema this is no problemo

jtaaa (commented):
Ideally I'd want to rename all the enum types in `generated/graphql` to be prefixed with `Gql` or something. Right now our shared and generated enums share the same name which can be a bit confusing.

I think we probably will rarely need to import the enum type from generated but if we do we can import it `as` `GqlEnumName`

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
chriskrogh: What is the type of `tierData?.store?.tier` is it a `string` or a `number`?

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
chriskrogh: Does sc snd server have different behaviour for using `shared` enums?

packages/shared/src/enums/Tier.ts
```diff
@@ -0,0 +1,5 @@
+export enum Tier {
+  Free = '0',
+  Launch = '1',
+  Growth = '2',
+}
```
chriskrogh, line 5: This will infer `number` types not so?

```suggestion
export enum Tier {
  Free,
  Launch,
  Growth,
}
```

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
jtaaa: It's an enum with keys and values both as the GraphQL enum member names.

`tierData?.store?.tier` is a key of the `shared` `Tier` enum (or undefined)

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
jtaaa: server uses only the `shared` enums internally whereas clients will use the GraphQL generated enum to index the `shared` enum for a value.

We are stuck with the generated enum since GraphQL has no notion of values for an enum so `apollo-client` just uses the enum member name as the value when we make requests.

packages/shared/src/enums/Tier.ts
jtaaa: Yessir and they should be numbers not string 😅  good catch I was trying stuff out and forgot to swap back

chriskrogh (approved):

PR #166, opened 2020-09-29
Use shared enums in client packages

jtaaa:
## What's changed

In most of the places throughout `seller-client` and `shared-components` I have replaced imports of `shared` enums from `generated/graphql` with imports from `@sellyourstuff/shared`.

## Notes

If a component takes an enum value as a prop it should probably use the enum from `shared`.

In some cases, for example in address, where the ServiceArea enum is nested in the address, it is easier to take the whole address and convert the enum in the JSX.


jtaaa:
Task linked: [CU-cwtc5y Client-side enums are broken](https://app.clickup.com/t/cwtc5y)

chriskrogh (approved):

PR #167, opened 2020-09-29
AddAddress improvements

jtaaa:
## What's changed

The `townCity` field in the `AddressForm` now has label "Area" and it is a `MuiSelect` when the country is set to Trinidad and Tobago. The `country` field is now a `MuiSelect` as well with the only two options Trinidad and Tobago and USA.
The `type` field is new and is either `Residential` or `Commercial`.

## UI

Screen Recording 2020-09-29 at 12 06 48 PM

## Notes

In Storybooks we have an extra option for the `type` field. "AddressType" shows up as a selectable value. I think it's because Storybooks uses [react-docgen](https://github.com/reactjs/react-docgen) which is adding the `displayName` property to our enum.


jtaaa:
Task linked: [CU-bwjg79 AddAddress improvements](https://app.clickup.com/t/bwjg79)

chriskrogh (approved):

PR #168, opened 2020-09-30
ProductForm Submission - New product and Update product

jtaaa:
This ones a little bit ⚓ier

## What's changed

We can now submit the product form for new products and to update existing products 🚀 

## UI

Screen Recording 2020-09-30 at 3 30 53 AM

## Notes

Still need to add a `ProductStatus` status field for `active`, `inactive`, etc. Then groups.


jtaaa:
Task linked: [CU-cwrptm ProductForm submission](https://app.clickup.com/t/cwrptm)

jtaaa:
Task linked: [CU-cwtc5y Client-side enums are broken](https://app.clickup.com/t/cwtc5y)

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
```diff
@@ -86,17 +100,41 @@ const ProductSubPage: React.FC<ProductSubPageProps> = ({ close, id }) => {
...
+              onSubmit={onSubmit}
+              onNameChanged={setName}
+              uploadImages={uploadImages}
+              defaultValues={product}
+              tags={tags}
+              addTag={addTag}
+              tier={Tier[gqlTier ?? 'Free']}
+            />
+          )
+        }
```
chriskrogh, line 137: Can we use the ternary operator here and add 1 comment instead of 2? 
```suggestion
        { !product ? (
            <ProductForm
              formRef={formRef}
              onSubmit={onSubmit}
              onNameChanged={setName}
              uploadImages={uploadImages}
              tags={tags}
              addTag={addTag}
              tier={Tier[gqlTier ?? 'Free']}
            />
          ) : (
            <ProductForm
              formRef={formRef}
              onSubmit={onSubmit}
              onNameChanged={setName}
              uploadImages={uploadImages}
              defaultValues={product}
              tags={tags}
              addTag={addTag}
              tier={Tier[gqlTier ?? 'Free']}
            />
          )}
```

chriskrogh (approved):
LGTM!
Just one suggested change

packages/seller-client/src/pages/ProductsPage/SubPages/Product/ProductSubPage.tsx
jtaaa: I actually tried that and for sanity I loaded the branch and tried again.

I think React is too smart and knows that the only difference between the two trees when we `!product` changes is the props applied to `ProductForm`. So it doesn't bother to unmount the first `ProductForm` and mount the second `ProductForm`. It just updates the props on the component which causes a problem since `defaultValues` is not set on first render.

Interestingly I can't find any documentation about this anywhere but I have confirmed that a single `ProductForm` is being mounted once regardless of updates to `!product`.

I think a good solution to this would be to make `UpdateProductForm` and `NewProductForm` components that wrap `ProductForm` and only do 1 thing. Then they would be considered different components by React and the render problem would be fixed.

Another solution would be to stop rendering the empty `ProductForm` before the product is loaded and display a spinner or something instead. Not a big fan of this.

jtaaa: Having the two `ProductForms` as separate expressions makes React consider the two `ProductForm`s different even though the render trees would be really similar. Maybe because react considers the two expressions (ProductForms) as different nodes.

chriskrogh: https://reactjs.org/docs/conditional-rendering.html

Both ways only load one `ProductForm` right? I was suggesting the change for readability rather than performance.

jtaaa: Nope the current way loads two separate `ProductForm`s! And it mounts and unmounts each as expected.

With the ternary operator only one `ProductForm` is ever mounted and its props are changed when the ternary operator condition changes

chriskrogh: Ah that's what I thought initially. Why do we want to mount 2 `ProductForm`s as opposed to 1?

jtaaa: So that when the second one (product is defined) is mounted it will have `defaultValues` set on initial render and correctly set those defaultValues in `useForm`. After the initial render updating `defaultValues` does not change the default values of the form.

chriskrogh: Ah okay I see. Makes sense. It's kind of interesting that the change in `defaultValues` doesn't trigger a re-render. This is fine I guess!

jtaaa: Actually it does trigger a re-render! It just doesn't update the state of the form since the default values of the state have already be set.

chriskrogh: wow. wellI learned something new there. I thought a re-render with new props will change the default state of the component :/ Well that's !cool @ all

PR #207, opened 2020-10-13
Remote dev db and some small fixes

chriskrogh:
Might be useful to review by commit. Included a few small fixes and additions as well

## What's changed (In order of commit)
1. 50af44468f7f9992203ae670c17115f9bc81d82d
I added the `--inspect` flag to our `start:dev` script so that we can attach a debugger listener. This allows us to set breakpoints in `vscode` while our server is running. 
2. 0da3b1336e57980e89be3696cfb1ac8781f307d8
I removed the `unique` check for `facebookId` and `googleId` for 2 reasons
(a). they are both optional fields, do if one user has `null`, no other user can have `null`
(b). we can enforce uniqueness using emails instead
3. 32e4bbc42046e5cc9da34cc0c5b70d38c10cd43c
This was a small nit. I didn't like seeing "1 Options"😂
4. 2ba53abfbbfeff758ec4b095c8c3eb6465dd7f40
Adding the remote dev url🥳

## Notes

To use the remote dev url, you need to copy the contents of the `.env.production` into the `.env`.
To access the dev user account, log into the app using:
`email: test@user.com`
`password: Password1!`

This way if we onboard any more engineers, we can easily have them started up with some test data and they don't have to generate any locally.

## PS
I also added 1 more nit commit 2d09f1d348843d049a0b27ad152954c51504e640. it changes our **vscode** workspace package names to remove the `@sys...` Our **yarn** workspace package names will still be prefixed with `@sys...` to specify scope, however developers don't have to see those names wile using vscode because the scope is quite obvious. But for imports we want to be as explicit as possible. lmk what u think
Screen Shot 2020-10-13 at 11 55 11 AM


packages/server/.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+  "configurations": [
+    {
+      "name": "Attach",
+      "port": 9229,
+      "request": "attach",
+      "skipFiles": ["<node_internals>/**"],
+      "type": "pwa-node"
+    }
+  ]
+}
```
jtaaa, line 15: 😍

packages/server/src/utils/constants.ts
```diff
@@ -48,6 +50,7 @@ export const PORT = ENVVARS.PORT;
 export const ROOT_URL = ENVVARS.ROOT_URL;
 export const SELLER_APP_URL = ENVVARS.SELLER_APP_URL;
 export const SELLER_STORYBOOK_URL = ENVVARS.SELLER_STORYBOOK_URL;
+export const USE_REMOTE_DEV_DB = ENVVARS.USE_REMOTE_DEV_DB;
```
jtaaa, line 53: what do you think about converting to a boolean here?
```suggestion
export const USE_REMOTE_DEV_DB = ENVVARS.USE_REMOTE_DEV_DB === 'true';
```

packages/server/src/utils/mongoose.ts
```diff
@@ -1,12 +1,21 @@
...
 
 const dbName = 'sys';
+const remoteUrl = `mongodb+srv://${MONGO_USER}:${MONGO_PW}@${MONGO_HOST}/${dbName}`;
 const mongoUrl =
   NODE_ENV === 'production'
-    ? `mongodb+srv://${MONGO_USER}:${MONGO_PW}@${MONGO_HOST}/${dbName}`
+    ? remoteUrl
+    : USE_REMOTE_DEV_DB === 'true'
+    ? `${remoteUrl}dev`
     : `mongodb://localhost/${dbName}`;
```
jtaaa, line 19: we can set `dbName` before like
```ts
const dbName = USE_REMOTE_DEV_DB ? 'sysdev' : 'sys';
```
then we don't need the extra ternary operator when setting mongoUrl

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
```diff
@@ -33,7 +33,9 @@ const ProductVariant: React.FC<ProductVariantProps> = ({
         </Typography>
         <Spacer width={16} />
         <Typography variant="caption">
-          <strong>{optionsCount} Options</strong>
+          <strong>
+            {optionsCount} Option{optionsCount === 1 ? '' : 's'}
```
jtaaa, line 37: I bet this functionality will be needed elsewhere!

jtaaa (commented):
Looks so good!

I'm down for renaming the workspaces in vscode. I originally set them to match the yarn package names but we don't need the scope in vscode

packages/server/src/utils/constants.ts
chriskrogh: I've been trying with this for a while, and for some reason I can't get our `ENVVARS` values to be typed as `string | boolean`, they really wanna be strings. Also that change is a bit involved and can probably be its own PR as a chore.
I'm fine with leaving it as a `string` for now and refactoring later because we have similar code like this for one of the other env vars as well. What do you think?

chriskrogh: https://app.clickup.com/t/dwpak3

chriskrogh: Okay so hear this right. I swear I tried this out and it didn't work.
But then I tried it again and it works :-)

chriskrogh: Will update🤣

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
chriskrogh: I agree. But I feel like this is one of the things that if we extract, people will forget to use and then it just becomes unused. 
The engineer writing the code will have to remember to call the function if we extract it, and if they remember to call the function, then they can remember to add the logic for a possible 's'? 
Do you think people will remember to call it?

packages/server/src/utils/mongoose.ts
chriskrogh: updated in 8d12d7b0072bff5b5605c72006bf89fed9bfcda0

packages/server/src/utils/constants.ts
chriskrogh: updated in 70a084731b93ae650b9b862bb9d208ceec4d9479

chriskrogh:
reintroduced unique and added sparse in a569f68bc73771c8214eb73a2c5ae5012aa3ef15

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
jtaaa: I would remember to call it! Definitely doesn't need to be extracted but it could help readability

jtaaa (approved):
🚀

packages/shared-components/src/components/ProductListItem/ProductVariant.tsx
chriskrogh: alright let's have em!
185f68362e5124ba70bbf88abd94b32ddb6ea161

PR #209, opened 2020-10-13
Added storybook for page header

chriskrogh:
[CU](https://app.clickup.com/t/bjka5c)

## What's changed

The ticket wanted to add the header for the order subpage so that we can display the order code, but we can just fo this through the page title instead.

## UI

Just title
Screen Shot 2020-10-13 at 12 27 58 PM

With buttons
Screen Shot 2020-10-13 at 12 28 53 PM


jtaaa (approved):
🦜

PR #210, opened 2020-10-13
Orders in Search Bar

chriskrogh:
Kind of hard to read :(
Better to review by commit

## What's changed
1. fa699071f94c2a1371df4835acfa3e07b74e4075
The search results type had a cyclical definition where it depended on the prop types of the `ProductListItem` component. I corrected this by setting the type as `ProductSearchResultFragment` which is the actual type of the data that we expect to receive.
2. 269dd37f1721d6ef629e76ec453e5709c82cd8e5
I updated the names of the props to `OrderListItem` to match the db model `Order` (makes passing props easier). I also noticed that `expectedDeliveryDate` can be `null`, so we need to make the the `date` prop optional to handle that case.
3. bdf6805ed98a9a2df69844f966d19b8379a7879a
Added orders to the search bar and some types using fragments.

This is kind of a WIP PR. I couldn't test the search for orders from storybooks because I actually have to be authenticated to query for a store's orders. I'm following this up with the `OrdersPage` PR to integrate it (needed the fragments from this PR)


packages/seller-client/src/components/SearchBar/SearchBar.tsx
```diff
@@ -66,12 +67,28 @@ const SearchBar: React.FC<SearchBarProps> = ({ include }) => {
...
+      case SearchGroup.Orders:
+        return (
+          <Box width="100%">
+            <OrderListItem
+              {...option}
+              date={
+                option.expectedDeliveryDate
+                  ? new Date(option.expectedDeliveryDate)
+                  : undefined
+              }
```
jtaaa, line 79: Small nit. I'm a fan of moving the logic out of the JSX. Like defining `const date` above the `return`. Not a big deal since this is pretty small and easy enough to read in the JSX

packages/seller-client/src/components/SearchBar/SearchBar.tsx
```diff
@@ -136,7 +153,8 @@ const SearchBar: React.FC<SearchBarProps> = ({ include }) => {
   );
 
   // use inputValue as option label when name is undefined to always match input
-  const getOptionLabel = (option: SearchOption) => option.name ?? inputValue;
+  const getOptionLabel = (option: SearchOption) =>
+    (option.searchGroup === SearchGroup.Products && option.name) || inputValue;
```
jtaaa, line 157: I think we will need a switch statement here to return order `code` for orders and `name` for products. with `inputValue` as fallback

jtaaa (approved):
Looks good!

packages/seller-client/src/components/SearchBar/SearchBar.tsx
jtaaa: @chriskrogh this guy is important!

chriskrogh: Will update before merging!

chriskrogh: d0e1c26dac40a5520f1f4decda75a30d41f21d08

packages/seller-client/src/components/SearchBar/SearchBar.tsx
chriskrogh: d0e1c26dac40a5520f1f4decda75a30d41f21d08

PR #211, opened 2020-10-13
Added Orders Page

chriskrogh:
## What's changed

Added the orders page.

## UI

Screen Shot 2020-10-13 at 3 00 11 PM

## Notes

I'm going to fix up the remote dev db first so that I can test with some orders that I created over there. (Too lazy to make some more orders locally)


packages/server/src/resolvers/Order/OrderResolver.ts
```diff
@@ -51,12 +51,12 @@ class OrderResolver {
       .exec();
   }
 
-  @Query((_returns) => Order)
+  @Query((_returns) => [Order])
```
jtaaa, line 54: 😭  my b!

jtaaa (approved):
Looking great! looking forward to seeing it working with data

packages/server/src/resolvers/Order/OrderResolver.ts
chriskrogh: NP!

PR #213, opened 2020-10-14
Set up GitHub Actions to build and test

jtaaa:
## What's changed

Use GitHub Actions for CI/CD


.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Triggers the workflow on push
+on: [push, pull_request]
+
+jobs:
+  # This workflow contains a single job called "build"
```
chriskrogh, line 9: this is prob an old comment

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Cache node_modules folder and bust cache on yarn.lock changes
```
chriskrogh, line 53: this comment is also a lil misleading because we're pulling from the cache instead of caching? 
```suggestion
      # Get node_modules folder from cache and update dependencies if yarn.lock has changed
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+        with:
+          node-version: 12.x
+
+      - name: Download shared build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: shared-build
+          path: packages/shared/dist
+
+      - run: yarn install --frozen-lockfile
```
chriskrogh, line 72: how come we pull node_modules from cache and then run this after?
If we're gonna update deps, we shouldn't cache node_modules

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+      - run: yarn workspace @sellyourstuff/server generate
+      - run: yarn workspace @sellyourstuff/shared-components build
+
+      - name: Upload GraphQL schema artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: schema
+          path: packages/server/schema.gql
```
chriskrogh, line 80: very nice😍

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Cache node_modules folder and bust cache on yarn.lock changes
```
chriskrogh, line 99: ^

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+          name: shared-build
+          path: packages/shared/dist
+
+      - name: Download shared-components build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: shared-components-build
+          path: packages/shared-components/dist
+
+      - run: yarn install --frozen-lockfile
```
chriskrogh, line 130: ^

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Cache node_modules folder and bust cache on yarn.lock changes
```
chriskrogh, line 150: ^

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+          name: shared-components-build
+          path: packages/shared-components/dist
+
+      - name: Download marketplace build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: shared-components-build
+          path: packages/shared-components/dist
+
+      - run: yarn install --frozen-lockfile
```
chriskrogh, line 181: ^

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Cache node_modules folder and bust cache on yarn.lock changes
```
chriskrogh, line 201: ^

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
...
+          name: shared-build
+          path: packages/shared/dist
+
+      - name: Download shared-components build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: shared-components-build
+          path: packages/shared-components/dist
+
+      - run: yarn install --frozen-lockfile
```
chriskrogh, line 232: ^

chriskrogh (commented):
Do we have a circular dependency between shared-components and server for the `schema.graphql`? Wouldn't the server have to generate a new schema if the models change for ex? This could break other things

.github/workflows/build.yml
jtaaa: true this probably isn't needed!

sidenote:
It currently doesn't actually update the deps since they already exist from the cache and presumably all have the correct versions

.github/workflows/build.yml
jtaaa: damn yeah I forgot to clean the comments

jtaaa:
> Do we have a circular dependency between shared-components and server for the `schema.graphql`? Wouldn't the server have to generate a new schema if the models change for ex? This could break other things

We generate the schema in the shared-components-build using `yarn workspaces @sellyourstuff/server generate` so there shouldn't be any circular dep

.github/workflows/build.yml
```diff
@@ -0,0 +1,240 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Triggers the workflow on push
+on: [push, pull_request]
```
chriskrogh, line 6: Also how do you feel about only having this run a `push`? [push] encapsulates [push, pull_request]?

jtaaa: I think pull_request is actually slightly different. At least it is in TravisCI where it willrun the job on the result of merging the PR into the target branch

jtaaa: I'll check to see what the difference is in GitHub actions too

chriskrogh: If it's anything like ⭕️ then it'll do one for each. It looks like it in the pic above^

jtaaa: what I'm saying is I think the `(pull_request)` boys are actually running the job on different code than the `(push)` boys. Gotta do more investigation tho. Found this fat [blog post](https://frontside.com/blog/2020-05-26-github-actions-pull_request/#:~:text=Note%20that%20git%20checkout%20is,be%20refs%2Fheads%2Fbranch_name%20.) that I think will hold the answers. Will go through it soon!

jtaaa: Actually I skipped to near the bottom and found this
> When you use pull_request, @actions/checkout will perform a git checkout to the github.ref environment variable. Note that git checkout is not applied to the commit, as it would have been the case when using push.

> This difference means that a pull_request workflow ref would look like refs/remotes/pull/##/merge whereas a push workflow would be refs/heads/branch_name. This explains why the SHA of a push workflow matches the commit that triggered the workflow, whereas the SHA of a pull_request workflow does not; instead the SHA of the pull_request is the resulting commit that was created from merging the base to the head.

So [push] doesn't necessarily encapsulate [push, pull_request]

jtaaa: I'm down to maybe disable `push` on pull requests but enable it in the master branch? I think the job run on `pull_request` is the more useful/safe option

.github/workflows/build.yml
jtaaa: Took another look and we actually do need that line!

If the `yarn.lock` has changes and the cache misses then we don't have any `node_modules` installed and we need to install with that line.
If the cache hits then yarn is smart enough to know that everything is up to date and the line does nothing.

I'm going to remove the `yarn install`s for all the other jobs since this first one should build the cache. Should be safe 😅  and it'll save 5-7s each job 🐎

.github/workflows/build.yml
jtaaa: I implemented run on push for the master branch and run on all pull requests [`70e1884 `](https://github.com/SYSTT/sys-monorepo/pull/213/commits/70e1884f145d0b7b4e97d62dcc72c59d9ecdd5e9)

Let me know what you think!

chriskrogh: SGTM!

.github/workflows/build.yml
chriskrogh: Yeaaa sg!

chriskrogh (approved):
>perfect! looking so good!
>How soon do you think we can disable Travis and make our repo private?

I hope as soon as #217 gets merged! Hopefully very very soon. I've poured way too many hours into this

jtaaa:
> perfect! looking so good!
> How soon do you think we can disable Travis and make our repo private?

I hope as soon as #217 gets merged! Hopefully very very soon. I've poured way too many hours into this

PR #216, opened 2020-10-14
WIP Order Page

chriskrogh:
I realized that there were so many small changes that I might as well make a PR for them now before the diffs get ⚓️

## What's changed
1. d34d94890198e0fea3014cd6c850c974006372f7 and d1c3ea867b05b69564cb8e025ecf8c07778aa184
I realized that the order variant of `ProductListItem` didn't have a way of showing quantity, so I added it. We also pass this from the `OrderDetails` component so that needed an update as well.
2. 35d8902d485f0fcf2b208d02bb5c91899c9994b6 and 22f86774b9aae494072d6ebffe336a2833d9bdd6
Apparently the plural of status is status. Keeping that in mind makes things readable, however someone who isn't aware of this may be confused. Let me know if this is confusing and you want to revert to `statuses` I kinda respect English tho.
3. 1470f84729299d9f910727c2769b3a6e893b3af6
Added paid out status
4. cb999bffa39eae8046805ac55343d81ff4cc3718
Added Order query
5. 6b5a1b026a8d36ced207230d77ed36f48e1d90f9
Denormalized `image` and `productId` from product onto option
6. daef0586d2f5339dfcde2d2d363f1930e5c72e22
Renamed `address` prop to `deliveryAddress` to match the model. Also since it's optional, I had to make the address optional in the buyer details. I also made ratings optional for now in d0ea965cb3886f78c91dbd270f2bf0fa617700c6 because we don't have them in release 1.

without addr
Screen Shot 2020-10-14 at 6 16 12 PM
with addr
Screen Shot 2020-10-14 at 6 16 07 PM


jtaaa:
Task linked: [CU-drum1v Order SubPage](https://app.clickup.com/t/drum1v)

packages/server/src/resolvers/Order/OrderResolver.ts
```diff
@@ -61,6 +61,13 @@ class OrderResolver {
     return await OrderModel.find({ store: store._id }).exec();
   }
 
+  @Query((_type) => Order)
+  async order(
+    @Arg('id', (_type) => ObjectIdScalar) id: ObjectId,
+  ): Promise<Order | null> {
+    return await OrderModel.findById(id).populate('items.productOption.image');
```
jtaaa, line 68: you should be able to use
```suggestion
    return await OrderModel.getFromUserInputId(id).populate('items.productOption.image');
```

and it will handle throwing with a nice error if the id does not correspond to an existing order

packages/server/src/resolvers/Product/ProductResolver.ts
```diff
@@ -84,7 +84,9 @@ class ProductResolver {
       optionsData.map((optionData) => ({
         ...optionData,
         product: product._id,
+        productId: product._id,
```
jtaaa, line 87: Instead of actually denormalizing the data in the DB we can instead define a custom field resolver in the ProductOptionResolver that will return `productOption.product`. We don't need to denormalize in the DB since we already have the id in the `product` field as a ref.

Something like

```
  @FieldResolver((_type) => ObjectIdScalar)
  productId(@Root() productOption: ProductOption): ObjectId {
    return productOption.product;
  }
```

as a toplevel member of the resolver

jtaaa (commented):
>35d8902 and 22f8677
>Apparently the plural of status is status. Keeping that in mind makes things readable, however someone who isn't aware of this may be confused. Let me know if this is confusing and you want to revert to statuses I kinda respect English tho.

I much prefer statuses since it gives indication to the reader that it is multiple vs being ambiguous. I personally don't respect English that much to live the with ambiguity. If you feel strongly about it I'm open to hear why!

>df576b4
>We copy the delivery address onto the order (if delivery is !pickup) so that if the buyer changes their delivery address, the order will still have a copy of it.

Superb catch!

packages/server/src/resolvers/Order/OrderResolver.ts
jtaaa: or you may have to call the populate like this
```ts
const order = await OrderModel.getFromUserInputId(id);
await order.populate('items.productOption.image');
```

packages/server/src/resolvers/Product/ProductResolver.ts
jtaaa: shouldn't be too complex and would save us from storing the `product._id` on the productOption twice

jtaaa:
Everything looks good to me so far! I have a few suggestions above.

I'm super grateful you're able to figure out and sort out some of the things I've overlooked like the delivery address information on an order and even defining `deliveryAddress` as a GraphQL field! I hope that these little mistakes on my part aren't slowing you down/frustrating. I'll try my best not to introduce any more of these little issues (and now we have code-reviews too so that should help)

jtaaa (approved):
I think I was a bit confused because I label PRs with `[WIP]` when they aren't ready to be put through yet! That's why I didn't approve this or the other.

I'm approving so that I don't slow you down but I think the DB normalization stuff might be worth sorting out one time.

PR #217, opened 2020-10-15
GitHub Actions deploy

jtaaa:
## What's changed

🚧 👷🏽‍♀️ 🚧 
- [x] Deploy the master branch on every new push through GitHub actions 🚀 
- [x] Rip out Travis ⚰️


jtaaa:
I'm so confident that it's going to work this time that I'm going to start ripping out Travis.

Thx for the builds buddy @Travis

jtaaa:
ofc the deploy still isn't working 🙂

packages/server/scripts/install-and-restart.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
chriskrogh, line 1: Why do we have a script to install and restart?

jtaaa: Our GitHub Actions deploy job will download all the build artifacts and transfer them to the server.

We need to install the node_modules on the server in case any deps have changed - we don't transfer the node_modules folder because it's several hundred MB

Then we just restart the pm2 process - we can probably setup `pm2` to watch the files and restart for itself but I haven't looking into that as yet!

chriskrogh (approved):
Great work bro!
41 fckin commits eh JEES
🚀

jtaaa:
I'm gonna merge this guy with admin privs since Travis doesn't wanna let go just yet 😄

PR #224, opened 2020-10-15
More WIP 

chriskrogh:
## What's changed
1. d08b134cd68a6c04f108b9e2e210985345656597
updated default value of the total label to be "Total" instead of "Your Total".
2. d24e1040ee0563f8f5aca46064e52551cee78676
I noticed that in checkout and in orders, we include the delivery fee if it exists, so I updated the shared component to accommodate for the delivery fee.
3. 79b2978716404792a0f090d496a2bfa95241ad13
I added a small loading component. You can use it like a page when loading a single product or order by passing a `height` prop of "90vh" or so. But you can also use it as a centered component by passing a `marginTop`

## Notes

Next PR will hopefully be my last sc PR for a lil while!


jtaaa:
Task linked: [CU-drum1v Order SubPage](https://app.clickup.com/t/drum1v)

packages/shared-components/src/components/Loading/Loading.tsx
```diff
@@ -0,0 +1,24 @@
+import type { BoxProps } from '@material-ui/core/Box';
+
+import React from 'react';
+import { Box, CircularProgress } from '@material-ui/core';
+
+const Loading: React.FC<Pick<BoxProps, 'height' | 'marginTop'>> = ({
+  height,
+  marginTop,
```
jtaaa, line 8: What do you think about a boolean `center` prop?

jtaaa (commented):
Looks good to me!

packages/shared-components/src/components/Loading/Loading.tsx
chriskrogh: I'm down!

jtaaa:
Want me to approve this?

chriskrogh:
> Want me to approve this?

If it lgty then sure! I guess I will have some rebases when I address the PR feedback from https://github.com/SYSTT/sys-monorepo/pull/216 but after that we should be gtg

jtaaa (approved):

PR #227, opened 2020-10-16
El Finito

chriskrogh:
## What's changed

Added the Order SubPage

## UI

Screen-Recording-2020-10-15-at-1 (2)

Screen-Recording-2020-10-15-at-1


jtaaa:
😍 😍 
gifs are looking great

will take a look soon!

updating the cache would also ensure other (theoretical) components that rely on the status would get the update

jtaaa (approved):
Looks super clean bro ¡¡

